PR #48, opened 2026-07-17
[MERGE #47 FIRST] Enterprise audit: close every codeable finding + remove SQLite + fix the gate

Ranjith36963:
Closes the codeable findings from the `docs/fable/` enterprise audit, removes SQLite entirely, and fixes two correctness holes in the commit gate itself.

**Every commit is full-suite-green** (final: 1,761 backend + 166 frontend).

## P0 / P1
- **DB connection pool** (`15c5b68`) — per-request connection; the shared psycopg singleton was neither concurrency-safe nor self-healing
- **Transactional migrations** (`ea18e10`) — atomic BEGIN/COMMIT. Root cause was subtle: the shim's speculative `SELECT lastval()` aborted the transaction, so a first attempt was reverted rather than shipped broken
- **Real account erasure** — `hard_delete_user` across 17 tables (closed 3 findings at once)
- Sentry-in-worker · auth-page error leak · PostHog funnel · magic-link + client-log

## P2
- **Security** — Redis rate limits (gated, atomic Lua, in-memory fallback) · timing-safe login · defusedxml · CSRF (incl. converting a side-effecting GET to POST) · lockout email+IP
- **Data** — purge orphan cleanup · ISO timestamps · normalized_key · identity-sequence resync · atomic Kanban moves
- **Compliance** — consent gate (PostHog never loads without consent) · GDPR Article-20 export · DB audit trail (migration 0025) · Dependabot
- **Ops/Frontend/Harness** — railway.json · Node 22 · fail-closed middleware · a11y · gitleaks

## SQLite removed entirely
The app was already 100% Postgres at runtime, but the code *wore* SQLite: helpers imported under SQLite aliases, a conftest `sys.modules` swap, a dead `aiosqlite` dependency, and 8 dev scripts importing real stdlib `sqlite3` (broken since the migration). That disguise directly caused three of the bugs above. **Zero SQLite imports remain.** The SQL *dialect* rewrite (`translate()`) is flagged in CLAUDE.md as the follow-up batch.

## Score Radar — it displayed a model that never existed
Git-verified: the radar shipped with 8 **invented** axes, then the DB was bent to match the UI rather than the engine. Result: a perfect title match plotted at **267%**, three axes were permanently 0, and salary/visa/workplace were computed every request and **discarded**. Fixed by pointing the chart at the real engine (which happens to have exactly 8 dims). `dims_active` is explicit — never inferred from zeros, because a 0 can be a real *earned* score.

## The commit gate had two correctness holes
- **The stamp bound the wrong tree** — fingerprinted *after* the 13-min run, so edits made during a run were blessed by tests that never covered them. Now start/end verified.
- Plus a per-database advisory lock (kill-safe) and a heartbeat, after an incident where silence was misread as death.

## Docs
36 findings now carry their own **STATUS + commit** — the finding docs previously described everything as open, which had already cost another session real work. A PII-in-logs finding that **this audit missed** (caught externally) is recorded *as a miss*, with why.

## Left open — deliberately
Owner decisions, not defects: **SI1** (notifications need a Railway worker + Redis + creds — no code fixes it), privacy/terms + subprocessor list, scraping ToS, MFA. Dev-only paths (D7/D8/D12, F2, O8) left as accepted trade-offs. PR-2 (dropping 4 dead columns) **declined** by the owner — all-zero and harmless vs irreversible DDL on live data.


Ranjith36963:
## ⚠️ DEPENDENCY: merge #47 BEFORE this PR

This branch's history contains `3f532b2`, the stale-base commit that clobbered 4,348 lines in PR #44. **This PR does not cause that clobber, and does not re-apply it — but it does not fix it either.**

**Merging this PR alone leaves `main` missing:**
- `.github/workflows/security.yml` — your bandit / gitleaks / pip-audit CI (95 lines)
- `frontend/src/lib/security-headers.ts` — the M15 fix
- `backend/tests/test_xxe_hardening.py`, `test_security_hardening.py`, `test_pg_translate.py`
- `docs/FABLE_FINDINGS.md` (438 lines)

**#47 restores all of it** → https://github.com/Ranjith36963/job360/pull/47 (titled "MERGE FIRST" by its author).

### Verified, not assumed
- **This PR deletes nothing.** The merge base with #47 *is* `3f532b2` itself, which already lacks those files — so relative to the base, this branch removes nothing. `git merge-tree` confirms it never touches `security.yml`. Once #47 lands, the files survive this merge.
- **Not a live security hole.** `grep defusedxml` returns 0 on main, but main is still protected: `_sanitize_xml()` in `backend/src/sources/base.py` strips `<!DOCTYPE>`/`<!ENTITY>` before parsing, and billion-laughs needs an entity declaration. Verified present on this branch too (absence of *a* fix ≠ absence of protection).

### Expect conflicts — both sessions edited these
`migrations/runner.py` · `repositories/pg.py` · `repositories/pgsync.py` · `api/main.py` · `api/routes/{auth,jobs,tailor}.py` · `services/auth/email_sender.py` · `workers/tasks.py` · `tests/{test_api,test_account_mgmt,test_api_idor}.py`

Not resolved here **deliberately**: #47 is another session's branch, and the original clobber was caused by exactly this — one session resolving another's work across worktrees. Whoever merges should take #47 first, then resolve against it.

PR #90, opened 2026-07-18
fix(H7): make pip-audit blocking with 11 documented aiohttp CVE waivers

Ranjith36963:
Closes the **security-scanner half of H7**. Stacked on #89 (contains its commits).

## The 11 CVEs were one stale pin

| | Before | After |
|---|---|---|
| `aiohttp` | 3.13.5 (`>=3.9.0,<3.14`) | `>=3.14.1` |
| `aioresponses` | `>=0.7.0` | `>=0.7.9` |

`PYSEC-2026-237`, `-2104`..`-2113` — 11 CVEs, all in `aiohttp`, all fixed in 3.14.x.

The `<3.14` pin was **not** a compatibility floor for our own code. It existed because aiohttp 3.14 broke `aioresponses` 0.7.8 — the offline HTTP mock every source test depends on (rule #4). `aioresponses` 0.7.9 declares `aiohttp>=3.8,<4.0` and supports 3.14, so the reason for the pin is gone.

With pip-audit's only findings resolved, it flips to blocking. **All four scanners now fail the build on a finding.**

## Verification — please read, the local run does NOT prove the bump

Local gate: full backend suite, **1,936 passed / 3 skipped, 27m54s**.

That run used **aiohttp 3.13.5** — the version already installed. It proves this change breaks nothing at the *current* version. It does **not** prove 3.14.1 works, and it cannot: this machine's interpreter is shared with unrelated projects, so installing 3.14.1 there would disturb them. (A local `pip-audit` there is equally useless — it reports CVEs from `torch`/`yt-dlp`/`tornado` that have nothing to do with this repo. That is exactly why every number in this PR comes from CI's logs.)

**CI is the verifier that matters.** It installs from `pyproject.toml` clean, so it resolves aiohttp 3.14.1 + aioresponses 0.7.9 and runs the whole suite against them. If 3.14 still breaks the mock, the offline suite fails loudly and this does not merge. That is the intended check, not a gap.

## Side effect worth knowing about

The shared test Postgres had **115 leaked `t_*`/`mem_*` schemas** from previously killed runs. Every test rescans `information_schema`, so the suite had degraded from ~30 min to a projected **~9 hours**. Dropped them (after confirming 0 active connections); the suite is back to ~28 min.

## H7 after this

- Scanners: **done** — all four blocking
- `mypy`: still non-blocking, 803 strict-mode errors (not 395 — the doc is stale). Ratchet coming in the next PR.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -3,27 +3,22 @@ name: Security scanning
 # H7 — continuous security scanning: dependency CVEs (backend + frontend),
 # leaked secrets, and Python static analysis. Runs on every push/PR and weekly.
 #
-# H7 follow-up (2026-07-18): THREE OF FOUR JOBS ARE NOW BLOCKING. They were all
-# advisory "until the codebase is clean"; nobody came back to check whether that
-# day had arrived. It had — verified against a clean CI runner (run 29648821735,
-# PR #88), which is the only trustworthy source here (a local pip-audit reads
-# whatever else shares your interpreter):
-#   npm audit  — "found 0 vulnerabilities"   -> BLOCKING
-#   gitleaks   — "no leaks found"            -> BLOCKING
-#   bandit     — "No issues identified" (22,619 lines) -> BLOCKING
-#   pip-audit  — 11 CVEs, ALL in aiohttp 3.13.5 -> still advisory, see below
-#
-# pip-audit is the one real gap, and it is a single stale pin: aiohttp is held at
-# <3.14 because 3.14 broke aioresponses 0.7.8 (our offline HTTP mock, rule #4).
-# aioresponses 0.7.9 supports aiohttp 3.14, so the bump closes all 11 CVEs. That
-# is deliberately a SEPARATE change — a dependency bump must be verified by the
-# full suite against the NEW version, which only CI can do (it installs clean;
-# a local run tests whatever is already installed, i.e. the old version).
-# Flip pip-audit to blocking in the same PR as that bump.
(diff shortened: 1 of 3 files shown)


Ranjith36963:
## Course correction — the bump does not work

The original approach in this PR (bump `aiohttp` to >=3.14.1) **fails**. CI measured what I had only inferred:

```
TypeError: ClientResponse.__init__() missing 1 required
  keyword-only argument: 'stream_writer'
  → aioresponses/core.py:197, in build_response
```

aiohttp 3.14 made `stream_writer` a **required** kwarg of `ClientResponse.__init__`; `aioresponses` never passes it. 0.7.9 is the latest release, so **no version of the mock works with aiohttp 3.14 today**.

I had reasoned from `aioresponses` declaring `aiohttp>=3.8,<4.0`. That is a claim, not a test — and it was wrong.

Worth noting: **`pip-audit` went green** on the bump, because the CVEs genuinely are fixed in 3.14.1. Only the *test suite* failed. A scanner alone could never have caught this.

## What this PR does now

Keeps `aiohttp<3.14` and waives the 11 CVEs **by ID**, with pip-audit **still blocking**:

- new CVE in any other package → **build fails**
- a *new* aiohttp CVE → **build fails**
- only these 11 documented IDs → pass

Strictly stronger than the `|| true` it replaces, which hid every finding including these 11. The waiver list should only ever shrink — remove it when `aioresponses` supports aiohttp 3.14.

Gate: 1,936 passed / 3 skipped, 20m21s.

PR #130, opened 2026-07-26
fix(ci): audit gate blocks on production deps, reports dev-tooling advisories

Ranjith36963:
A `brace-expansion` DoS advisory (**17 vulns, 14 high**) is blocking **every open PR**, including #129.

## Why it can't just be patched
| Attempt | Result |
|---|---|
| Pin each line to its newest release (1.1.16 / 2.1.2 / 5.0.8) | audit **still** flags them — the advisory range is a flat `<=5.0.7`, so only 5.0.8 counts as patched |
| Blanket override to `^5.0.8` | audit clears ✅ — **but eslint breaks** ❌ (verified: `@eslint/config-array` crashes in `pathMatches`) |

**The advisory cannot be cleared without breaking the toolchain.**

## The fact that resolves it
`npm audit --omit=dev` → **0 high**, verified against main's **untouched** dependencies.

All 14 highs are **devDependencies** — eslint, vitest, openapi-typescript. They run on CI runners and dev laptops. **They are not in the shipped Next.js bundle.** No user can reach them.

## The change — one file, zero dependency edits
- **Blocking gate** → `npm audit --omit=dev --audit-level=high`. A high in code that *ships to users* still stops the build.
- **Second step, non-blocking** → still reports dev-tooling advisories for visibility.

## Why no lockfile changes (learned the hard way here)
This PR originally pinned `brace-expansion` via overrides. Reverted, because **main's own lockfile fails `npm ci` on Windows** (`@emnapi/wasi-threads` — a platform-specific optional dep) while passing on Linux CI. Lockfile edits **cannot be validated on this machine**, and shipping unverifiable ones broke this PR's CI twice. Pinning is deferred to an environment where `npm ci` can actually verify it.

Net diff: **`security.yml` only.**

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -85,9 +85,30 @@ jobs:
       - name: Install frontend deps
         working-directory: frontend
         run: npm ci
-      - name: Audit frontend dependencies (high+)
+      # BLOCKING gate — production dependencies only. These are the packages
+      # that actually ship in the Next.js bundle and run in front of users, so a
+      # high/critical here must stop the build.
+      - name: Audit PRODUCTION dependencies (blocking, high+)
         working-directory: frontend
-        run: npm audit --audit-level=high
+        run: npm audit --omit=dev --audit-level=high
+
+      # NON-BLOCKING — dev/build tooling (eslint, vitest, openapi-typescript).
+      # These never reach a user: they run on this runner and on a developer's
+      # laptop, and are not part of the deployed bundle. They were made blocking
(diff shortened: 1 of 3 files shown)


Ranjith36963:
Superseded by #133, which now carries **both** changes. They deadlocked: #133 ran the old gate (blocked by dev-tooling highs) and #130 ran the new gate (blocked by the prod shadcn→MCP→brace-expansion chain that #133 fixes). Neither could merge alone. Closing in favour of the combined PR.

PR #132, opened 2026-07-27
fix(security): patch brace-expansion DoS in prod deps; split the unfixable dev half

Ranjith36963:
## Problem
`GHSA-mh99-v99m-4gvg` (**brace-expansion: DoS via unbounded expansion**) turned the **npm audit (frontend deps)** job red on every PR. All **14 HIGH** findings traced to this one package. Pre-existing on `main` — not introduced by any branch.

## Production is genuinely fixed
`@sentry/bundler-plugin-core` and `@ts-morph/common` pulled `brace-expansion@5.0.7` via `minimatch`. Fixed with a **version-scoped** override in `frontend/package.json`:

```json
"brace-expansion@^5.0.0": "^5.0.8"
```

`npm audit --omit=dev --audit-level=high` → **exit 0**.

## The dev half has no upstream fix — so the gate is split, not waived
`5.0.8` is the **only** patched release; there is no 1.x/2.x patch. Our 1.x copy comes from **`eslint-plugin-jsx-a11y@6.10.2` (the latest release)**, which pins `minimatch ^3.1.2`. minimatch 3 does `require('brace-expansion')` expecting a bare function, so forcing 5.x on it kills ESLint entirely.

Both blanket fixes were tried and **both broke lint**:

| Attempt | Result |
|---|---|
| override `brace-expansion` → `^5.0.8` everywhere | ❌ `TypeError: expand is not a function` |
| override `minimatch` → `^10.2.5` everywhere | ❌ `_minimatch.default is not a function` |
| `npm audit fix --force` | ❌ breaking majors (eslint 10, shadcn 3) |

So `security.yml` now has two steps:
- **Production deps — BLOCKING** (the shipping surface, audited strictly)
- **Dev tooling — informational**, with a comment naming exactly what blocks the fix and **what would unblock it** (jsx-a11y shipping `minimatch >=10`, or a backported 1.x/2.x patch)

This is deliberately *not* a `continue-on-error` on the real gate. Blocking on an unfixable dev-tool CVE pins CI permanently red — which is how this repo previously ended up ignoring a scanner sitting on 11 real CVEs (see the workflow's own header).

## Verification
| Check | Result |
|---|---|
| `npm audit --omit=dev --audit-level=high` | exit 0 ✅ |
| `npm run lint` | exit 0 ✅ |
| `npm run type-check` | exit 0 ✅ |
| Gate (frontend suite) | 210 passed ✅ |

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -9,6 +9,10 @@ name: Security scanning
 # which is the only trustworthy source here (a local pip-audit reads whatever
 # else shares your interpreter):
 #   npm audit  — "found 0 vulnerabilities"
+#     UPDATE (2026-07-30): GHSA-mh99-v99m-4gvg (brace-expansion DoS) landed and
+#     turned this red. Production deps are FIXED (override -> brace-expansion
+#     5.0.8) and still blocking; the dev-tooling half has no upstream patch and
+#     is split into a separate informational step. Full reasoning at that step.
 #   gitleaks   — "no leaks found"
 #   bandit     — "No issues identified" (22,619 lines)
 #   pip-audit  — was 11 CVEs, ALL in aiohttp 3.13.5. Now ZERO: aiohttp is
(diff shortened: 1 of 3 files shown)


Ranjith36963:
Superseded by #133, which landed the same audit split (`--omit=dev` blocking + non-blocking dev step) and additionally moved `shadcn` to devDependencies. Verified: main's Security scanning run at `1884dd5` is green, so the `brace-expansion` override here is no longer needed. Closing rather than resolving the conflict — the work is already on main.

PR #225, opened 2026-08-05
feat(feed): user-feedback loop — liked/applied protected, not_interested excluded from selection

Ranjith36963:
## The user-feedback loop — USER-understanding brick

The action buttons (`liked` / `applied` / `not_interested`) wrote rows the candidate layer completely ignored: a rejected job could ride the feed forever; a liked/applied job could be silently evicted by re-selection. **The explicit user signal was a dead lever.**

Now, in both backfill and profile-change rescore:
- **`not_interested`** → excluded from selection, never reaches the paid LLM judge, force-evicted past every guard (including the LLM-verdict protection — the user's NO outranks the machine's opinion).
- **`liked`/`applied`** → join the kept set: never staled by re-selection, reactivated if stale. User investment survives, same principle as the verdict guard.

Note: the diff vs main is only the feedback-loop change (`f53a355`); earlier commits in the list are already on main via #224's squash.

**Tests:** +2 (rejected-excluded, liked-survives); 40 passed across candidate/rescore/feed suites. Gate green.

### DIFF
--- a/backend/scripts/embed_catalog.py
+++ b/backend/scripts/embed_catalog.py
@@ -1,0 +1,111 @@
+"""Backfill embeddings for every catalog job that doesn't have one.
+
+Job-understanding goal (2026-08-05): the semantic engine (E3) can only rank
+what is embedded, and prod had 284 of 7,309 jobs embedded (3.9%) — so hybrid
+retrieval silently degraded to keyword-only. This script closes that gap:
+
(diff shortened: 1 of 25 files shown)


Ranjith36963:
Superseded by the clean cherry-pick branch (pre-squash history conflicted with main).

PR #293, opened 2026-08-11
fix(profile): the CV merge dropped two shelves — make it derive, not hand-list

Ranjith36963:
Production caught this, not the tests.

PR #292 added `cv_experience_level` and `cv_right_to_work` — taught the prompt, declared them on `CVSchema`, surfaced them in the adapter, wired them to the judge, bumped `EXTRACTOR_VERSION`. All green. A real re-extraction on the live profile then produced **both shelves empty**.

`_merge_cv_llm_into` named every scalar it copied, one `if not cv.x and llm_cv.x` per field. The two new ones weren't in that list, so the LLM extracted them and the merge discarded them one layer above the fix.

**Third time this function has lost a field that way.** The comment being replaced documents `career_domain` suffering it on 2026-08-07, and the LinkedIn merger in #292 was the same shape again. Adding two more lines would have set up the fourth.

So the scalar merge now derives from the dataclass: every plain-string `CVData` field is CV-owned unless it carries a `linkedin_`/`github_`/`about_me_` prefix (another pass owns it) or appears in `_NOT_CV_OWNED_SCALARS` with a stated reason. Eight fields today; one added tomorrow is carried without touching this function.

## Why the tests missed it

They set `cv_experience_level` on the CVData **directly** and asserted downstream behaviour, so the merge was never on the path. The new tests assert structurally: whatever the CV pass produces on a CV-owned scalar must reach the profile, with no list of names to keep in step — plus that other passes' scalars are never clobbered by a CV re-parse.

## Verification

850 tests across the 38 files that touch `CVData` or any renamed/added symbol, **selected by symbol rather than by diff** — a diff-based selection is exactly what let the previous miss reach CI.


Ranjith36963:
Superseded by a clean branch off main — #292 was squash-merged, leaving this branch conflicting and CI unable to run.

PR #319, opened 2026-08-12
fix(rule30): delete the -15 foreign penalty and the hand-typed city list

Ranjith36963:
Owner-approved. Rule #30 says UK-only is a **door**, not a penalty, and never hand-enumerate an unbounded set. The code violated all three "never" clauses.

## What the list was actually doing

```
379 of 9,196 live rows (4.1%) were being docked 15 points
  ...including 9 UK jobs docked BY ACCIDENT
  "Belfast, Northern Ireland"  matched the list entry  "ireland"
```

Genuinely UK roles were being pushed down the feed by a list that exists to keep foreign roles out. That is rule #30 proving itself: an unbounded set cannot be hand-typed, so it rots and mis-fires.

## Test-first, and the tests were watched failing

Every new assertion failed by exactly 15 points before the fix — `Bangalore, India` 43 vs 58, `New York, US` 43 vs 58, `London, Ontario` 53 vs 68. The first draft of one test was **rigged** (London-Ontario earns +10 via "London") and the paired assertion caught it. 8 tests that pinned the banned behaviour were deleted.

Removing the penalty also **exposed a real hole in the door**: the dual-site escape accepted a UK gazetteer hit from the foreign segment itself — the UK has hamlets called California and New York — so `San Jose, California, US` walked in as a two-site ad. Fixed, with 4 tests watched failing.

## A BLOCKER was found in this branch and fixed

The first rewrite of the fetch filter asked the gate about whatever string it was handed. Four callers hand it an ad **description**, and the gate splits on `/`:

```
"</li>"  ->  segment "li"

li, br, td, tr, th, hr  are ALL in foreign_admin.txt
                        Liechtenstein, Brazil, Chad, Turkey, Thailand, Croatia

=> every HTML job ad read as a foreign country and would have been
   dropped AT FETCH, before the door, with no log line
```

Narrowed to bare location strings; markup and prose go straight to the door. New `test_fetch_filter_never_overreaches.py` pins the invariant that makes a pre-filter legitimate at all: **anything this drops, `check_uk` must also refuse.** It does not test the filter's reasoning — it tests that the filter never disagrees with the chokepoint in the dangerous direction, which is the assertion that would have caught the HTML bug on run one.

## Numbers

`SCORER_VERSION` 3 -> 4, so frozen `user_feed` rows re-score. 48 live rows' worth of clearly American addresses now stop at ingestion instead of being admitted-then-docked.

Known and documented, not hidden: foreign jobs that still slip the door (the London-Ontario class) now rank *higher* than before, because the second filter is gone.

test_sources + test_uk_gate + test_scorer: **272 passed**. New guard: 26 passed. Gate: PASS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UK-location eligibility checks, including foreign addresses and dual-site postings.
  * Ensured foreign jobs are rejected consistently before scoring.
  * Removed the foreign-location score deduction; location scoring now uses standard location matching.

* **Documentation**
  * Updated scoring, search, provider, and UK eligibility guidance to reflect the revised behaviour.

* **Tests**
  * Added regression coverage for UK, remote, foreign, and multi-location postings.


Ranjith36963:
@coderabbitai full review

PR #320, opened 2026-08-12
fix(rescore): move re-scoring onto the worker queue, add a resumable backfill (#271)

Ranjith36963:
Owner-approved. Closes the root cause behind #271.

## The bug

A profile save fired `rescore_user_feed` with `asyncio.create_task` **in the FastAPI process**. No queue entry, no retry, no completion record — `grep -c rescore src/workers/tasks.py` was **0**. `main` auto-deploys on every merge, so a deploy alone killed it silently.

```
9,708 feed rows on a stale profile_version
      0 of them orphaned  <- every one was reachable and simply never re-scored
  1 user at profile version 10 while their current version is 15
```

## Part A — the queue

`rescore_user_feed_task` in `workers/tasks.py`, registered in `WorkerSettings.functions`, raising arq `Retry` with 30s×job_try backoff under the existing `max_tries=5`. New `workers/queue.py` is a shared, never-raising `enqueue_job()` door. If Redis is unreachable it falls back to the old in-process task and logs `QUEUE UNAVAILABLE` — both paths tested.

## Part B — the backfill, where the danger is

This repo's own recorded lesson: *a correctness fix with an operational cost is still a regression* — a previous backfill froze the event loop, and 5-row tests could not see it. So it is batched, throttled, **resumable** (dies at row 5,000 → next run continues, no double-work), and is a job you run deliberately, not on boot.

## 15 tests, all watched failing first

Not "written then green" — every one produced an ImportError or AssertionError before any implementation existed. Several were **mutation-tested**: making the job id include a `uuid4` produced `the job id changed between runs`; forcing `queued = False` produced `the profile save did not enqueue`; deleting the fallback broke the dead-Redis test.

The load-bearing one is `test_the_backfill_lets_other_work_run_between_batches` — a competing coroutine runs alongside the backfill and must get scheduled. That is the event-loop-starvation threat, tested directly.

## A HIGH was found in this branch and fixed

Only `create_pool` was bounded. arq maps `conn_timeout` to the **connect phase only**, so a Redis that accepts the TCP connection and then stops responding left `await pool.enqueue_job(...)` waiting forever — hanging a user-facing profile save **and** never reaching the fallback, so the work was lost. Both calls are now wrapped.

**Honest gap:** that specific timeout is verified by reading the code path, not by an executed test. I wrote one and it passed for the *wrong reason* — a real dev Redis is running here, my monkeypatch did not take effect, and the call simply succeeded against the live instance. A test that goes green without exercising what it names is the exact defect this session spent the day removing, so I deleted it rather than ship it.

## What you need to do after merging

1. **Confirm the API service has `REDIS_URL` and the worker is up.** Otherwise every profile save quietly takes the in-process fallback and nothing improves. Look for `rescore: queued a durable re-score for user ...` (good) vs `rescore: QUEUE UNAVAILABLE` (bad).
2. **Run the backfill deliberately**, once you are happy, to clear the 9,708 existing rows.

Gate: PASS.

## Summary by CodeRabbit

* **New Features**
  * Profile updates now trigger reliable background feed re-scoring when relevant content changes.
  * Feed re-scoring automatically retries after temporary failures and processes affected accounts in batches.
  * Added a command to launch and control large-scale feed re-scoring operations, including batch size, limits and throttling.

* **Bug Fixes**
  * Profile saves remain successful when background queue services are temporarily unavailable.
  * Re-scoring can fall back to in-process processing when queue services cannot be reached.


Ranjith36963:
@coderabbitai full review

PR #321, opened 2026-08-12
fix(job_signals): the third instance of one packaging bug — signals go live

Ranjith36963:
Owner-approved. Third and final instance of the same bug.

```
1st  ESCO ............ never built at all              (documented, PR #305)
2nd  uk_gazetteer .... never in the image              (fixed, PR #312)
3rd  job_signals ..... never in the image              (THIS PR)
```

`job_signals._DATA` resolved `parent.parent.parent/"data"/"job_signals"` — fine on a dev disk, `<site-packages>/data/job_signals` in the container. `_load_terms()` returned `{}` for all three vocabularies, **silently**, so `signal_backed_lookup` filled nothing and `detect_seniority` quietly answered nothing.

Fixed by copying #312's proven pattern: data moved inside the package, declared in `[tool.setuptools.package-data]`, path resolves from site-packages **and** `/app`, and `_load_terms` now logs an ERROR naming the path when a file is missing or empty.

## Measured against your live catalog (9,196 jobs, read-only)

```
seniority coverage      27.6%  ->  57.9%      2,787 jobs get a real tier
                                              senior 2,047 · director 516 · junior 151
workplace_type          758 jobs get a real value
```

## Three corrections the verifier forced — read these

**1. Scores move BOTH ways.** I told you signals going live would improve scores. That was wrong. `scoring_dimensions.py:151-168`: UNKNOWN scores a neutral half-weight (+4), but a *detected mismatch* scores **−8**. So roughly half of the 2,787 newly-typed jobs may rank **DOWN** by up to 15 points. The post-deploy check must measure movement in both directions, not just coverage.

**2. The visible chips do not change yet.** `seniority` and `workplace_type` on a job card are read straight off stored DB columns, which this change never writes (`jobs.py:180-192`). Only the per-dimension **scores** move. The chips change when the enrichment pipeline next writes those columns.

**3. The mechanism claim was too broad.** Both Dockerfiles do `pip install .` **and** `COPY . .`, producing two copies of `src`. uvicorn puts cwd first on `sys.path`; arq puts it last. So the API resolved `/app/src` and the **worker** resolved site-packages — the bug bit them differently, not identically. Comments corrected.

## Tests

Watched failing first, including the nastier half — a file that exists but parses to nothing, where `path.exists()` is True and the vocabulary is still empty. `test_shipped_data.py` rewritten to check the RIGHT mechanism per location (`src/data/...` → pyproject package-data; `data/...` → dockerignore exemption) instead of one rule for both.

Proof that matters: **the wheel was built and its contents listed** to show the data files are inside. A test passing from the working tree proves nothing about the image — that is what caused all three instances.


Ranjith36963:
@coderabbitai full review

PR #322, opened 2026-08-12
fix(pillar1): eight defects a 12-agent audit found, six of them mine from today

Ranjith36963:
A full adversarial audit of the user side: six lenses, each finding refuted by a separate agent. **33 of 35 findings survived.** The most useful ones proved *me* wrong.

## The shrink was not a bug, and my guard was the wrong fix

I reported that a re-extraction shrank `job_titles` 3→2 and `education` 6→3, and guessed `load_profile` was duplicating on read. **Three lenses independently disproved that** against the real 10-version production history: `load_profile` is a pure deserialiser. The "6" is a **mid-run peak** — `enrich_cv_from_linkedin` re-appends LinkedIn's wordings, then `llm_merge_duplicates` collapses genuine near-duplicates back down. That is the feature working.

My refuse-to-write-on-shrink guard would have frozen *"Master of Science…"* and *"Master's degree, …"* into the profile as two separate degrees forever. **It is not shipped.**

The real defect underneath is different and worse: `llm_merge_duplicates` had **no lower bound**. Its only checks were "not empty" and "not longer than the input", so any shorter answer was accepted and written. A rate-limited model returning one item from seven silently deleted six certifications. It now must prove **coverage** — every entry it removes must still be represented by something it kept. Every existing test patched this function out, so the one step that can shorten three matching shelves had **no coverage at all**.

## Two data-loss bugs

| | |
|---|---|
| **Re-uploading a CV** deleted LinkedIn's education, certifications and job titles | Those three lists are written by *both* passes (LinkedIn has no shelf of its own for them), but the reset dropped only the `cv` cache key. The LinkedIn pass became a cache hit, contributed nothing, and the data was **unrecoverable by re-uploading the same PDF** — its text is unchanged, so the hash still matches. |
| **`linkedin_skills`** wiped on any unrelated profile edit | The 2026-08-08 `llm_ran` gate protected five sections. Skills have the same shape and the same failure, but the assignment sat **one line outside** the gate. On the owner's real profile that is 13 skills collapsing to 3. |

## Six were mine, from today

- `reset_cv_owned_fields` never cleared `cv_experience_level` / `cv_right_to_work` / `cv_projects` — uploading a **different person's CV** left the previous person's stated seniority and work authorisation. Confidently wrong is worse than missing. Now derived from the same source the merge uses.
- `embeddings` read key `title` from `cv_projects` while the CV pass writes `name` — projects reached the vector without their name *or* technology list.
- `test_every_matching_role_is_a_real_role` asserted `used & MATCHING_ROLES - MATCHING_ROLES`, which is `(used & M) - M` — **the empty set for every possible input.** It could not fail. Rewritten, and on its first real run it immediately caught the next item.
- **`feed` was a declared matching role while `feed.py` reads zero shelves** — pure inflation in the headline number.
- `cv_experience_level` and `cv_right_to_work` were read by the judge and shown nowhere — the same scored-on-but-invisible gap I fixed for `linkedin_summary` in #296 and **repeated one commit later**.

## "50 shelves feed matching" was misleading, so it's now computed

```
19  rank EVERY job      (scorer, search-keywords, tiering, prefilter)
22  reach only the judge (a capped window)
 9  reach only the vector
```

**19 is the number worth quoting.** A headline that cannot be recomputed from the code is exactly the kind of claim this batch keeps correcting.

## Region preferences still deleted the whole feed

#297 fixed the country case and stopped one level too high. Measured after it shipped:

```
United Kingdom  ->  keeps everything
West Midlands   ->  keeps NOTHING   (Birmingham is IN it)
Yorkshire       ->  keeps NOTHING   (Leeds is IN it)
```

The gazetteer answers this from **data**, not a typed list of regions (rule #30): it holds ~52k settlements and no admin divisions, so a preference absent from it is one this gate cannot evaluate — it passes instead of deleting. A settlement still filters normally. Fails open if the gazetteer is missing, which has already happened once.

Gate: PASS — 314 backend targeted, 253 frontend, api-types drift clean.

### DIFF
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,6 +1,31 @@
 # Runtime data — never bake into the image; mount as a volume.
 data/
 
+# ...EXCEPT the data that is not runtime state but SHIPPED REFERENCE DATA.
+#
+# `.gitignore` deliberately un-ignores these same two directories, which is the
+# whole reason they are committed. This file did not mirror that, and the
+# Dockerfile is `COPY . .`, so the build threw away exactly the files someone had
+# gone to the trouble of exempting — and nothing rebuilds them at deploy time.
+# Railway volumes start empty, so no mount could supply them either.
+#
+# What that cost, silently and with no error:
+#   uk_gazetteer/  — the ~52k UK place names rule #30 is built on. Without it
+#                    `uk_gate._read()` returns an empty frozenset (it degrades
+#                    rather than raising), so no job can match a UK place and the
+#                    gate falls through to source-trust and ad-text evidence.
+#   job_signals/   — the seniority / workplace / location term tables behind
+#                    `signal_backed_lookup`, which fills 'unknown' enrichment at
+#                    read time, and `detect_seniority`. `_load_terms` returns {}
+#                    on a missing file, so those detectors quietly answer nothing.
+#
+# Both failure modes are invisible: empty data, no exception, no log line.
+# `tests/test_shipped_data.py` now pins the two ignore files together.
+!data/uk_gazetteer/
+!data/uk_gazetteer/**
+!data/job_signals/
+!data/job_signals/**
+
 # Python caches
 __pycache__/
 *.pyc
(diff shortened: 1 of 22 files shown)


Ranjith36963:
Superseded by a clean branch off current main — this one was based on pre-squash commits, so GitHub saw it as CONFLICTING and no CI workflow fired.

PR #326, opened 2026-08-13
chore(ci): unblock the two loops a main ruleset would silently kill

Ranjith36963:
**Must merge BEFORE the ruleset goes active.** Both of these break silently the moment it does.

## 1. Granting a bypass actor is NOT enough

This inverts the assumption the whole ruleset plan rested on:

```
gh pr merge  ->  sees mergeStateStatus: BLOCKED
             ->  refuses CLIENT-SIDE
             ->  never calls the API
             ->  bypass_actors never gets a chance to apply
```

(cli/cli#13388, open against gh 2.90.0.) `doc-sync.yml:359` self-merges its standing PR every day. Without `--admin` that call starts failing and the PR piles up forever while the loop reports nothing wrong. Fixed.

## 2. pr-shepherd would create the deadlock it exists to clear

On a CONFLICTING PR it calls `update-branch` (`pr-shepherd.yml:84`), producing a commit authored by `github-actions[bot]`. A GITHUB_TOKEN-attributed push **parks** the resulting runs in `action_required` — they never start on their own.

`pr-repair.yml:410-418` already handles exactly this after its own push. pr-shepherd never did, so its summary line *"CI re-running"* was a claim, not a fact.

Today that is merely misleading. **With required checks on main it is fatal:** a parked run reports nothing, and a PR whose required checks never report can never merge. Copied pr-repair's approve loop in, and made the failure path say so loudly.

Both files re-parse as YAML; the modified pr-shepherd run block was extracted and checked with `bash -n`. Gate: PASS.

### DIFF
--- a/.github/workflows/doc-sync.yml
+++ b/.github/workflows/doc-sync.yml
@@ -356,7 +356,12 @@ jobs:
           else
             # GITHUB_TOKEN-created PRs trigger no CI, so --auto would wait forever.
             # The docs-only + no-destruction asserts above ARE the gate.
-            gh pr merge "${pr:-$branch}" --squash --delete-branch
+            # --admin is REQUIRED once main has a ruleset. `gh pr merge` refuses
+            # CLIENT-SIDE on mergeStateStatus=BLOCKED and never calls the API, so
+            # the bypass_actors entry never gets a chance to apply (cli/cli#13388).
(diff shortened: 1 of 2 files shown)


Ranjith36963:
@coderabbitai full review

PR #327, opened 2026-08-13
fix(synthetic): kill the 30-day cookie treadmill — the workflow logs itself in

Ranjith36963:
**Before:** every 30 days, log in, open devtools, copy the raw `job360_session` cookie, paste it into a repo secret. Miss it and synthetic-live goes blind — which already happened for 3 days while 9 authed corners reported OK against a login page.

A recurring manual chore attached to a safety check is a check with a countdown on it.

**After:** the workflow logs in at run time. `POST /api/auth/login` (`auth.py:215-276`) takes email + password and returns the session cookie in one request — not magic-link only, so no mailbox access is needed anywhere. Set two things once, delete `SMOKE_SESSION`, never touch it again.

## Safety is in code, not in a comment — two independent gates

```
1. BEFORE the password leaves the process
     the address must be a dedicated synthetic one
     -> a human's password is never transmitted, no session ever minted
     verified live: ZERO network calls made on refusal

2. AFTER a session exists (by login OR by pasted cookie)
     ask the BACKEND whose account it is  ->  GET /api/auth/me
     refuse the walk unless the server's own answer is synthetic
```

Gate 2 is the one that matters: it does not trust the input, it asks the server. A pasted cookie for a real user fails it too.

## What you do, once

```
1. curl -X POST https://<backend>/api/auth/register \
     -H "Content-Type: application/json" \
     -d "{\\"email\\":\\"synthetic@job360.uk\\",\\"password\\":\\"<long random>\\"}"
2. gh variable set SMOKE_EMAIL     # the address
3. gh secret   set SMOKE_PASSWORD  # the password
4. delete the SMOKE_SESSION secret
```

Nothing expires on a clock after that.

### DIFF
--- a/.github/workflows/synthetic-live.yml
+++ b/.github/workflows/synthetic-live.yml
@@ -4,14 +4,68 @@
 # human and, on any failure, screenshots it + logs the REASON into the server log system
 # (POST /api/client-log). Fails red → GitHub emails the owner (scheduled-workflow default).
 #
-# Public corners always run. Authed corners run only when the SMOKE_SESSION secret is set
-# to a VERIFIED synthetic account's job360_session cookie (prod email-verification can't be
-# automated in CI). See tests/synthetic/live-smoke.mjs.
+# Public corners always run. Authed corners run only after the robot LOGS ITSELF IN as a
+# dedicated synthetic account and the backend confirms whose account that is.
+#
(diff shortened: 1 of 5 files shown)


Ranjith36963:
@coderabbitai full review

PR #328, opened 2026-08-13
fix(llm): a missing key can no longer look like an accuracy regression

Ranjith36963:
The weekly ranking eval reported *"audit produced no judgments — treating as failure"* for a **week**. The truth was that no judge key was set. A missing credential was reported as a **quality regression** — the most expensive kind of wrong alarm, because it sends you debugging your product instead of your config.

## One preflight where every path must cross

`llm_provider.require_llm_key()` now runs FIRST inside `llm_extract` and `llm_extract_fast` — it used to be a duplicated check AFTER the provider loop, so the real answer arrived buried under provider errors. It raises a dedicated `LLMKeyMissing`, names all four variables, and says what to do.

All 11 modules that touch the chain funnel through those two functions, so a caller written next month inherits it for free.

Masquerade sites closed: `llm_matcher.match_batch` no longer swallows it per job (N warnings + "judged 0/160" is what read as a quality problem), and `eval_ranking.py` drops its private copy of the four names.

## The detector this branch added was itself dead on arrival

Caught by adversarial review, and it is the same pattern this whole week has been about:

```
provider_probe.py    if not results: return 2      <- returns HERE
                     no_llm_key = ...              <- never reached
```

An env with no provider secrets — **exactly what your CI is** — returned 2 and never reached the alarm. Live proof: run `31683129750` (2026-08-13) exits 2, and `external-health` is `failure` on **8 of 8 runs since 2026-08-06**. It has never once been green.

So the promised *"within 24 hours an issue appears"* would never have happened. Fixed by computing it above the guard and reporting from there. Proven in the real condition, not a stub:

```
env -u OPENAI_API_KEY -u GEMINI_API_KEY -u GROQ_API_KEY -u CEREBRAS_API_KEY ...
  -> "## No LLM provider is configured at all — the judge is DOWN"
  -> exit 3        (was: exit 2, silent)
```

**Correction to the branch's own premise:** it claimed the probe had been GREEN daily while proving nothing. It was **RED daily and nobody read it** — a different failure mode, and one this PR does not solve.

## What you do, once

Set **any one** of `GROQ_API_KEY` / `GEMINI_API_KEY` / `CEREBRAS_API_KEY` (free tiers) or `OPENAI_API_KEY` (paid), as a repo Actions secret and on the Railway backend + worker. Not recurring.

### DIFF
--- a/.github/workflows/external-health.yml
+++ b/.github/workflows/external-health.yml
@@ -145,7 +145,10 @@ jobs:
           cat providers.md
           cat providers.md >> "$GITHUB_STEP_SUMMARY"
           echo "dead=$code" >> "$GITHUB_OUTPUT"
-          [ "$code" -le 1 ] || exit "$code" # exit 2 = nothing configured, fail LOUD
+          # 0 ok · 1 a key is dead · 3 a whole capability has NO key — 1 and 3
+          # are both reported as issues below, so the step itself stays green.
+          # 2 (the probe is blind) is the one that must fail the job LOUD.
+          case "$code" in 0 | 1 | 3) ;; *) exit "$code" ;; esac
 
       - name: Raise it, and hand it to triage
         if: always() && steps.probe.outputs.dead != ''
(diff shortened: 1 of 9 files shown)


Ranjith36963:
@coderabbitai full review

PR #329, opened 2026-08-13
fix(journey): the loop cannot run in CI — say all three reasons in one run

Ranjith36963:
`journey.yml` has been RED for 5 nights with one unhelpful line: `JOURNEY_API is not set`.

**Do not set it.** Investigating produced a verdict, not a value.

## Three blockers, not one

**1. The probe MUTATES, heavily.** `journey_probe.py` registers accounts (`:220`), uploads CVs (`:268`), uploads LinkedIn PDFs (`:305`), enriches from GitHub (`:341`), fires real searches that grow the shared catalog (`:359`), and runs raw SQL `UPDATE users SET email_verified_at = now()` (`:254-258`).

Pointing it at production = **7 junk accounts + 7 full source sweeps into the live database, every night, forever.** The workflow header already warned this; nobody had read it.

**2. There is no staging.** `railway status --json` -> `environments: ['production']`. Exactly one. Every backend that answers is prod (`backend-production-80e8e...` 200, `job360.uk` 200 — the latter is the frontend proxying to the same prod backend). **So the correct value does not exist yet.**

**3. The corpus is gitignored real-person CVs.** The probe needs `<corpus>/CV/*.pdf` (`:200-206`), defaulting to `User_info` — which `.gitignore:70` excludes and `profile_benchmark.py:14` calls *"PRIVACY — THE LOAD-BEARING RULE"*. A runner never has it and never should.

Setting the variable would have turned blocker 1 into blocker 2 and burned another night to learn it.

## The near-miss worth recording

```
gh secret   set JOURNEY_API   <- succeeds. changes nothing. WRONG NAMESPACE.
gh variable set JOURNEY_API   <- the real knob (journey.yml:74,92 read vars.*)
```

I gave you the wrong one earlier today. The preflight now names the right knob **and** the trap, so the next person does not repeat it.

## What this PR does

Reports all three preconditions in a single run, each with its exact command, instead of revealing them one wasted night at a time.

⚠️ **Adversarial review found a FOURTH:** `journey.yml:143` hardcodes `JOURNEY_VERIFY_DSN` to `localhost:5433` — a throwaway service container, not the target backend's database, and no step ever runs migrations against it. So even with a staging URL and a corpus, it stays red. Not fixed here; flagged because the header must not claim "two things must be true" when there are four.

## Your decision

This is not a chore. Either stand up a staging backend (costs money), or turn this loop off until you do — a nightly red that cannot go green trains you to ignore alarms.

## Summary by CodeRabbit

* **Bug Fixes**
  * Journey checks now clearly identify missing API target and corpus configuration.
  * Workflows fail early when either required prerequisite is unavailable.
  * Added guidance for correct setup commands, namespaces, corpus requirements, and staging/privacy considerations.

* **Tests**
  * Added coverage verifying configuration guidance and ensuring private corpus files are not tracked or committed.

### DIFF
--- a/.github/workflows/journey.yml
+++ b/.github/workflows/journey.yml
@@ -23,6 +23,15 @@ name: Journey probe (walk a real user, grade every step)
 #
 # NOT pointed at production by default. It CREATES accounts and RUNS searches,
 # so it belongs against a staging/ephemeral target. Set JOURNEY_API to opt in.
+#
+# TWO THINGS MUST BE TRUE BEFORE THIS CAN EVER BE GREEN, and neither is optional:
+#   1. vars.JOURNEY_API   — a backend BASE URL. A repo VARIABLE, not a secret;
+#      `gh secret set` writes the wrong namespace, succeeds, and does nothing.
+#   2. a corpus of CVs    — the default `User_info/` is gitignored real-person
+#      PII, so a runner checkout never has it. Set vars.JOURNEY_CORPUS to a
+#      committed synthetic corpus, or this loop is red no matter what #1 says.
+# The preflight below reports BOTH at once, on purpose: a guard that reveals one
+# blocker per night costs a night per blocker.
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


Ranjith36963:
@coderabbitai full review

PR #29, opened 2026-07-09
feat(email): production email readiness (PLAN-2) — visible failures + documented vars + tests

Ranjith36963:
## PLAN-2 — unblock real sign-ups (code half)

Job360 is live, but Resend is in **sandbox mode**, which only delivers magic-link login emails to the account owner. So no other human can log in. This PR is the code half; the actual unblock is a **human step** (below).

### Code changes
- **Failures are now visible.** The 4 delivery-failure logs in `email_sender.py` are `ERROR` (were `WARNING`). A broken provider (e.g. wrong `SMTP_FROM` domain → Resend 403) previously ate every login email silently while the user saw "check your email".
- **Documented the hidden env vars.** `RESEND_API_KEY` + `SMTP_FROM` were read only via `os.environ`, undocumented. Now declared in `settings.py` and `.env.example`. Removed the dead `SMTP_HOST`/`SMTP_PORT` constants (nothing imported them).
- **Tests for the untested failure paths:** `test_email_sender.py` (Resend 5xx + no-provider → returns False + logs ERROR), a magic-link send-failure test (still returns 204 — no email enumeration — and still mints the token so a resend works), and 3 frontend consume-page tests.

### 🔴 HUMAN STEP required to actually unblock sign-ups (I can't do this)
**Option A (recommended):** In Resend, verify a domain you own (add its DKIM/SPF DNS records). Then on Railway set `RESEND_API_KEY` (production key) and `SMTP_FROM=login@yourdomain`. Redeploy.
**Option B (stopgap, no domain):** Gmail app password → set `SMTP_EMAIL`/`SMTP_PASSWORD`/`SMTP_FROM`, and remove any `re_`-prefixed value from `SMTP_PASSWORD` (or the code takes the Resend path).
After either: if `REQUIRE_EMAIL_VERIFICATION=false` was set as a sandbox escape hatch, set it back to `true`. Then send a magic link to a NON-owner address and confirm login.

### Evidence
Gate: backend **1677 passed**, frontend **155 passed**, api-types no drift.

### Stacking
Based on `fix/ci-red-gate` (PLAN-1, PR #27). Merge #27 first, then this retargets to main.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #331, opened 2026-08-15
fix(pillar1): extraction never re-ran, and the form invented an answer

Ranjith36963:
Every defect here was **proven against production before a line was changed**, as asked. Three from the audit, two found while fixing them.

---

## 1. Extraction never re-runs — the big one

```
run_two_pass_extraction  →  ONE caller in src/:  profile.py:559
                            inside a save route a user must personally hit

worker crons (from the deployed worker's own startup line):
   nightly_ghost_sweep · refresh_catalog · notification_tick · enrichment_sweep
   ↑ none touch profiles

.github/ · migrations/ · scripts/  →  nothing
```

So bumping `EXTRACTOR_VERSION` for a prompt fix reaches only users who happen to re-save. **Everyone else stays frozen, silently, forever.** A live profile is stuck at extractor **v4** while the code is at **v6** — v5 and v6 have never run for it.

It has been papered over **four times** already. Production's version history records `shelf_backfill_reextract`, `github_enrich_repair` ×2, `backfill_linkedin_contact`, `backfill_linkedin_headline` — and `grep -rn` finds none of those strings anywhere in this repo. They were throwaway scripts.

Worse, they called `save_profile()` directly, stamping a fresh `updated_at` while leaving `llm_input_hashes` stale — **so a stale profile looks freshly updated.** That is exactly what made one audit lens misdiagnose this as a silently-failing LLM.

**Fix:** `two_pass.stale_extraction_inputs()` (reuses the existing digest, never reimplements it) plus a committed, dry-run-by-default sweep script with a `--limit` spend cap and a distinct `source_action` — a reviewable replacement for the four ghosts.

**Deliberately no cron.** Cadence and limit are a spend decision, not a mechanical one. Every re-extraction is a paid LLM call.

## 2. The form invented a seniority (rule #29)

A brand-new account — no profile row, `GET /profile` 404s, `page.tsx` passes `preferences={}` — posted `experience_level="mid"` the user never chose. Proven by a test that failed with `expected 'mid' to be ''`.

It is not inert: seniority scoring treats a typed value as authoritative (weight 8, ±8 on every job), and the same fake value went to the LLM judge and into the semantic vector.

The fix already existed **one field over** — `_normalize_work_arrangement`, from the "any" sentinel fix — and had never been copied across. Now mirrored. The route's default also changed from a bare `""` to the stored value, fixing a separate bug: any partial save silently wiped a stored `experience_level`.

## 3. The worker cannot embed, and said nothing

```
Dockerfile        (API)     torch + ".[semantic]"
Dockerfile.worker (worker)  pip install .          ← confirmed in the live build log
```

The worker runs `refresh_catalog`, so the process ingesting the most jobs is the one that cannot embed them — 687 of 9,977 (6.9%).

The failure shape is why it survived: `sentence_transformers` is imported lazily, so the module import **succeeds** and it only fails per job, inside the loop, where a per-job handler logs `embed backfill: job <id> failed`. One missing package produced a storm of identical warnings naming neither cause nor remedy.

**⚠️ This PR does NOT add the extra to the worker image.** That is real image size and build time — your call, not mine. What it does is make the failure **loud**: a preflight that logs one ERROR naming the cause and both ways out (install the extra, or turn `SEMANTIC_ENABLED` off so the gap is intentional rather than silent).

---

## Also fixed, found while fixing the above

| what | why it mattered |
|---|---|
| `reextract_stale_profiles.py` guarded its cap with `if apply and limit and ...` | `--limit 0` — the one invocation its own docstring promises "never makes an LLM call" — was **falsy**, so the cap was skipped entirely and it would re-extract every stale profile unbounded. Caught by the review pass. Pinned by a test I watched fail: `--limit 0 re-extracted 25 profiles` |
| `shelf_audit.py` had `two_pass` in `_WRITER_ROLES` but not `_ROLES` | Every read by the extraction orchestrator — the most shelf-heavy module on the user side — was invisible to `readers()`. It called `llm_input_hashes` read-only-by-"api" while `two_pass.py:220` compares its value to decide whether to skip a paid LLM call. A doc then believed the instrument. Matching count unchanged (49) |
| Completeness meter double-counted `target_job_titles` into two 15% buckets | A CV plus one typed title now reads **55%, not 70%** |
| `excluded_companies` collected and sent on every save | Backend has zero references to it. Removed |
| `test_the_allowlist_does_not_rot` iterated an empty dict | Its body never ran, so it could never fail. Rewritten, with a companion test proving the check fires against a fabricated entry |

## Verification

- 230 backend tests across the blast radius · 73 frontend profile tests
- ruff clean · tsc clean · eslint clean · `[gate] PASS`
- **Every new test was watched failing first.** Where an agent could not honestly claim that, it said so and I re-derived it

## What this does NOT fix

- **Sentry was never checked** in the audit — its MCP needs an OAuth round-trip no subagent could complete. No extraction-error signal has been examined.
- Whether `SEMANTIC_ENABLED` is on for the deployed services — `railway variables` is correctly permission-blocked, so the preflight above will answer it from the logs instead.
- **Coverage bound:** production holds two distinct CVs across three profiles, all driven by one person. Nothing here proves extraction works for a stranger.

## Summary by CodeRabbit

* **New Features**
  * Added a safe option to identify and reprocess profiles with outdated extraction results, including dry runs and processing limits.
  * Added a “No preference” option for experience-level selections.
  * AI model selections can now be configured through environment settings.

* **Bug Fixes**
  * Experience-level preferences now preserve omitted values, clear explicit empty selections, and reject unsupported entries.
  * Profile completeness scores no longer double-count selected job titles.
  * Preference autosaves no longer include excluded companies.
  * Semantic processing stops early with a clear setup error when required capabilities are unavailable.

### DIFF
--- a/backend/scripts/reextract_stale_profiles.py
+++ b/backend/scripts/reextract_stale_profiles.py
@@ -1,0 +1,149 @@
+"""Committed, reviewable replacement for FOUR throwaway re-extraction scripts.
+
+WHY THIS EXISTS. ``run_two_pass_extraction`` has exactly one caller in this
+codebase: a user-triggered save route, reached only when a user personally
+re-saves their profile. Nothing scheduled ever re-reads an EXISTING profile —
+so when ``EXTRACTOR_VERSION`` bumps because a prompt improved, every profile
+whose owner does not happen to re-save stays frozen at the OLD extraction,
+silently, forever.
+
+Production has already patched this FOUR separate times with uncommitted,
+throwaway scripts — their ``source_action`` values (``shelf_backfill_reextract``,
+``github_enrich_repair`` x2, ``backfill_linkedin_contact``,
+``backfill_linkedin_headline``) are visible in the live version history but
+none of those scripts, or those strings, exist anywhere in this repo. Worse:
+those scripts called ``save_profile()`` directly with a hand-patched field,
+which stamps a fresh ``updated_at`` while leaving ``llm_input_hashes``
+untouched — so the profile then LOOKS freshly updated while its extraction is
+still stale. That is what made a later investigation misdiagnose the cause.
+
+THIS SCRIPT is the honest replacement:
+  * Selects ONLY profiles ``two_pass.stale_extraction_inputs`` says are
+    actually behind — never a blanket re-extract. Each re-extraction is a
+    paid LLM call, so re-running everyone "just in case" is real money for
+    users whose extraction is already current.
+  * Re-extracts through ``run_two_pass_extraction`` — the SAME path a real
+    save takes — never a bespoke partial field patch. That bespoke-patch
+    pattern is the whole lesson of the four scripts this replaces.
+  * DRY-RUN BY DEFAULT. Prints what it would do; changes nothing without
+    ``--apply``.
+  * ``--limit N`` bounds how many STALE profiles get processed in one run —
+    the operator chooses how much to spend per invocation.
+  * Records a source_action distinct from every past ad-hoc value
+    ("stale_extractor_version_reextract"), so the version history says
+    plainly what touched the row.
+  * Safe to re-run: a profile already current (nothing in
+    ``stale_extraction_inputs``) is skipped, so running this twice in a row
+    is a no-op the second time.
+
+Follows the shape of ``scripts/cleanup_preference_pollution.py`` (dry-run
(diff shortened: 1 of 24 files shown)


Ranjith36963:
@coderabbitai full review

PR #332, opened 2026-08-15
chore(review): teach CodeRabbit this repo's actual failure mode

Ranjith36963:
CodeRabbit is installed and reviewing. Out of the box it looks for style and correctness. **Neither is what breaks here** — mypy is at zero, 2,700+ tests pass.

## What actually breaks here

Eight in one week, all measured:

```
a size check that swallowed its own exit code
a sed range whose anchor was deleted -> fed 82 wrong lines to the checker
a drift rule blind to the one file carrying its own bug
a route test reading a mutable global -> verdict changed with test ORDER
a Redis timeout covering connect but not the call
an LLM alarm sitting below the return that fires first
a freshness guard querying a column neither table has
an alert path whose missing key made it exit 0 GREEN for weeks
```

Every one was correct-looking code in the wrong position, reporting success about a **different question** than the one that mattered.

So the review question worth paying for is not *"is this correct"* but **"if this breaks at 3am, does anyone find out?"**

## Why a third reviewer

Claude writes most of this repo; Codex attacks it. Both are LLMs with overlapping instincts. Claude-reviewing-Claude corrected **30 of its own 69 verdicts** — good — but twice agreed with a confident wrong claim:

| I claimed | truth |
|---|---|
| "the pipe is severed" | starved — nobody searched |
| "quality was never measured" | measured 39% -> 66% -> 78%, then the key ran out |

A copy of you has no reason to doubt you. **Disagreement between models is the signal; agreement is weak evidence.**

## Config choices worth defending

- **assertive** — a polite third opinion is one he ignores, and this repo already died of ignored alarms
- **does NOT block merges** (`request_changes_workflow` + `fail_commit_status` both false) — `main` auto-deploys and he is solo. A third-party service holding the only gate is a single point of failure nobody can unblock at 2am. Enforcement lives in the branch ruleset where he holds the bypass
- **generated unit tests OFF** — a generated test is one nobody watched fail, which is exactly how eight guards shipped unable to fire
- **generated docstrings OFF** — docstrings here carry incident history; a generated one flattens that into a description of the signature

Six `path_instructions` encode the real incidents so it hunts the shape that keeps biting: SIGPIPE under `pipefail`, bare `failure()` gates that let one step lie about another, outputs only written on success, the two spellings of the bot identity, count-vs-freshness, silent-empty loaders, IDOR scoping, and doc claims citing files that do not exist.

Schema verified against `docs.coderabbit.ai/reference/yaml-template`, not written from memory. `tone_instructions` is 249 of 250 allowed chars.

**This PR is its own first test** — CodeRabbit will review the config that configures it.

## Summary by CodeRabbit

* **Chores**
  * Added comprehensive automated code review configuration with tailored guidance for workflows, scripts, API routes, services, tests and Markdown.
  * Enabled review summaries, status metadata, static-analysis checks and chat-based responses.
  * Configured automatic, advisory-only reviews with defined exclusions and repository-local learning scopes.
  * Standardised review language to British English.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,168 @@
+# CodeRabbit — the third reviewer.
+#
+# WHY THIS FILE EXISTS
+# Two reviewers already read every change here: Claude, which writes most of it, and
+# Codex/GPT, which attacks it. Both are LLMs with overlapping instincts. On 2026-08-11
+# Claude-reviewing-Claude corrected 30 of its own 69 verdicts — good, but it twice agreed
+# with a confident wrong claim ("the pipe is severed" when it was starved; "quality was
+# never measured" when it had measured 39%→66%→78%). A copy of you has no reason to doubt
+# you. This config exists to point a DIFFERENT reviewer at the failure this repo actually
+# has, rather than at style.
+#
+# THE FAILURE THIS REPO ACTUALLY HAS
+# Not bugs. Not types (mypy is at zero). Not tests (2,700+ pass).
+# It ships GUARDS THAT CANNOT FIRE. Eight of them, measured, in one week:
+#   · a size check that swallowed its own exit code
+#   · a sed range whose anchor had been deleted, silently feeding 82 wrong lines
+#   · a drift rule blind to the one file carrying the bug it was written for
+#   · a route test that read a mutable global, so its answer changed with test order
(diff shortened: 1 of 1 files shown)


Ranjith36963:
@coderabbitai full review

PR #347, opened 2026-08-17
feat(harness): the gate moves before the deploy — and the cage stops merging

Ranjith36963:
## The answer is no — and Anthropic's own tooling agrees

Their GitHub action, verbatim: *"Claude only creates and pushes commits. It does not merge branches."* Auto mode's classifier explicitly **blocks** "pushing directly to main, running production deploys." Their own Code Review *"always completes with a neutral conclusion so it never blocks merging."*

Nobody shipping this stuff lets the agent merge to production. That is a deliberate design choice, not an oversight.

## Measured, twice, five rounds

| | |
|---|---|
| PRs judged | **29** |
| allowed | **0** |
| agreement with owner's real merges | **0 / 18** |
| last 60 merges that would pass the path cage | **2 (3%)** |
| merges per week this lane would deliver | **zero** |

Converged at round 5 — verdicts stopped moving. Five blockers found and fixed along the way, **all of them wiring**, not judgement.

The refusals are not tunable. The dominant one is the owner's own sentence: his merge stream is product code and harness code, and those are the decisions he said stay his. Moving the number means contradicting the rule the cage exists to enforce.

## So the cage stops merging

`--merge` is removed. No file here merges anything. `dependabot-auto.yml` already covers the one class where unattended merging genuinely earns its keep — 5 of 245, all patch/minor, no model anywhere in it. Nothing is built beside it.

## The gate moves somewhere nobody looked

Not the PR. Not main. **The deploy trigger** — the gap between "merged" and "Railway starts building". It is the only place in this repo where a check that finishes late can still stop something.

Merge → live is 2m14s. CI lands ~4 minutes later. Railway's per-service **Wait for CI** flag closes exactly that window, covers 100% of merges, and writes no code.

## What survives

- `merge_cage --drill` **62/62**
- `cage_replay.py` — replays the cage over N merged PRs so the traffic number is re-measurable, not re-argued
- `--verdict-json` — per-cage verdicts, so nothing has to grep English to ask which cage refused
- `pr-advisor.yml` — comments on every PR naming **which file is the actual decision**. Merges nothing; worst case is a wrong comment. Covers 100% of PRs instead of the 3% a merge lane could reach.

## Two things the owner must know

**Two required contexts in ruleset 20793798 are decorative** — `Analyze (python)` and `Analyze (javascript-typescript)` name jobs, not reported check names. Enabling as-is would wedge the repo.

**There is no rollback.** Grepped every workflow and script: no scripted, documented, or drilled path back. That gap is worth more than any merge lane.


Ranjith36963:
Closing as **superseded by #348**, which merged today. The branch is untouched and this PR can be reopened.

## Why, with the evidence

**#347 and #348 were the same cage built twice.** Their `merge_cage.py` files were both 1,926 lines and differed by 3. `pr-advisor.yml`, `revert-main.yml`, `cage_replay.py`, `gate_wiring_check.py` and `revert_gear.py` were byte-identical on both branches.

They were never peers, though — #348 is the strict successor:

| | #347 | #348 |
|---|---|---|
| files | 20 | **192** |
| `.github/merge-policy.yml` | **absent** | present |
| `lane.py` (+409) | absent | present |
| `rollback_gear.py` (+653) | absent | present |
| docs/lane restructure | absent | present |

`git ls-tree origin/feat/automerge-that-earns-it .github/merge-policy.yml` is empty. #347 has no lane map at all.

## What it looks like against today's main

```
main + #347   conflicts=23   count-owed=FAIL   cage-drill=FAIL   wiring=FAIL
```

Its GitHub checks show green. They ran when its base branch still existed; #336 was squash-merged, so that history is gone from main. Green on a question nobody asked.

## What would be lost

Nothing. Everything #347 carries reached main through #348 — plus the lane map it never had. The four CodeRabbit fixes that lived only in the older lineage were ported deliberately in #357 (verified `grep -c 'required and concl'` returned 0 on #348 before porting, so "let #348 win" would have silently re-shipped a cage counting a skipped required check as a pass).

Reopen if you disagree — nothing here is deleted.

PR #349, opened 2026-08-19
fix(harness): chain_check reads heredoc writers — 18 live wires on #348 were reported dead

Ranjith36963:
## What this fixes

`Chain wires (harness)` fails on **PR #348** (`feat/two-lane-cage`) in 6 seconds with 18 broken wires. All 18 are in one file the cage adds, `.github/workflows/post-merge-watch.yml`, and **all 18 are false**. The workflow is correct. The checker cannot read it.

`scripts/chain_check.py` decided what a step writes to `$GITHUB_OUTPUT` with `_ANY_OUTPUT_KEY`, a regex that only matched lines beginning `echo` or `printf`. `post-merge-watch.yml` writes its outputs from a python heredoc instead:

```yaml
run: |
  python - <<'PY' >> "$GITHUB_OUTPUT"
  print(f"readyz_url={watch['readyz_url']}")
  PY
```

`print(...)` is neither `echo` nor `printf`, so `_parse_writes` returned `{}` and W1 reported all nine live wires out of that step as severed — *"it writes: nothing"*.

This is a **false negative in the instrument**, and it is the exact failure `chain_check.py` was written to catch: an instrument must count the way its consumer does. Left alone it is worse than the red — once someone silences it, the repo's post-merge prod watchdog ships with its wires permanently unchecked.

## Mechanism of the fix

Key discovery moves into one generator, `_iter_output_writes`, now shared by `_parse_writes` and `_write_depth` so the two cannot drift apart. It walks the `run:` block line by line, and when a heredoc's **opening line also redirects to `$GITHUB_OUTPUT`** it reads the body:

- python heredoc (`python`/`python3` in the opener) → `print(...)` / `sys.stdout.write(...)`
- plain heredoc (`cat <<EOF >> "$GITHUB_OUTPUT"`) → bare `key=` lines

Deliberate boundaries:

- A heredoc redirecting **anywhere else** is ignored, so a heredoc writing a temp file is never mistaken for one writing outputs.
- The `\s` before `<<` is load-bearing: it separates a real shell heredoc from the GHA multiline form `echo "body<<$d" >> "$GITHUB_OUTPUT"`, whose body is a *value*, not a script.
- Heredoc keys carry op `"<<"` — value unknowable — the same contract `key<<EOF` already had. So **W5 stays quiet rather than guessing** at literals it cannot evaluate.

## The drill — both directions, because only one of them is the trap

This repo's law is that a guard nobody has watched fail is indistinguishable from a guard that cannot fail. `Drill` gains `create` (write a *synthetic* workflow instead of mutating a real one — necessary because **no workflow on `main` uses this shape yet**) and `expect_silent`, plus an anti-vacuity guard: the synthetic file must actually parse and load, or the drill fails rather than passing on a file nobody read.

Two new drills pin both directions:

| drill | asserts |
|---|---|
| heredoc writer is READ, not mistaken for silence | a **correct** heredoc producer adds **zero** findings — the bug fixed here |
| heredoc writer with the wire actually cut | a heredoc producer with the wire **genuinely cut** still goes **RED as W1** |

The second is the one that matters. A "fix" that silenced this shape wholesale would clear the red *and* make the checker permanently blind to it. That drill fails if anyone ever does that.

## Measured (every line below is a command I ran)

```
main tree            python scripts/chain_check.py --no-map
                     -> 22 workflows, 21 wires, 0 broken, 1 suspect   exit 0
                     IDENTICAL before and after this change

cage tree (#348)     python scripts/chain_check.py --root <copy of cage .github> --no-map
                     before: 26 workflows, 22 wires, 18 broken, 1 suspect   exit 1
                     after : 26 workflows, 22 wires,  0 broken, 1 suspect   exit 0

counter-test         cut ONE real key in the cage's post-merge-watch.yml
                     (print poll_seconds= -> print poll_secondsX=)
                     -> 2 broken, exit 1, both W1 naming steps.parse.outputs.poll_seconds,
                        and it now lists all 9 keys it reads instead of "nothing"

drill                python scripts/chain_check.py --drill                    -> 6/6 passed, exit 0
meta-drill           python scripts/chain_check.py --drill --break-checker W1  -> 4/6, exit 1
                     (BOTH W1 drills fail when W1 is blinded — the new drill is load-bearing)

lint                 ruff: 7 findings before, 7 after — all pre-existing, none on changed lines
```

The CI log on #348 (`gh run view 32230340923 --log-failed`) prints `chain_check: 26 workflows, 22 handoff wires, 18 broken, 1 suspect` — byte-identical to the local reproduction, so the failure fixed here **is** the failure CI reports.

## Scope

One file, `scripts/chain_check.py`. Nothing under `backend/` or `frontend/` imports it (`grep -rn chain_check backend/` hits only `.ruff_cache` binaries), so no backend test is affected. CI runs `ruff`/`mypy` with `working-directory: backend`, so root `scripts/` is gated only by the two `Chain wires (harness)` steps in `ci.yml` — both run above and both pass.

---

# Landing plan for the harness branch stack (#336, #343–#348)

Not changed by this PR — recorded here because this PR removes one blocker from it. **Every merge below is a release: Railway auto-deploys `main` to job360.uk.**

State read live with `gh pr list`:

| PR | head | base | state | review |
|---|---|---|---|---|
| #336 | `feat/every-guard-declares-its-drill` | `main` | MERGEABLE / **CLEAN** | UNREVIEWED |
| #343 | `feat/the-sprawl-has-a-broom` | #336 | MERGEABLE / **CLEAN** | UNREVIEWED |
| #344 | `feat/main-can-finally-speak` | #336 | MERGEABLE / **CLEAN** | UNREVIEWED |
| #345 | `feat/the-gate-that-is-switched-off` | #336 | MERGEABLE / **CLEAN** | UNREVIEWED |
| #346 | `feat/findings-become-fixes` | #336 | MERGEABLE / **CLEAN** | UNREVIEWED |
| #347 | `feat/automerge-that-earns-it` | #336 | MERGEABLE / UNSTABLE | UNREVIEWED |
| #348 | `feat/two-lane-cage` | `main` | MERGEABLE / UNSTABLE | UNREVIEWED |

### Order

1. **#336 first.** It is the only source of `scripts/drill_registry.py`, which the cage references in four files (`merge_cage.py:345` shells out to it with `--count-owed`; also `lane.py:299`, `cage_blockers.py:425`, `data_only.py:9`). Expect zero conflicts.
2. **#343 → #344 → #345 → #346**, re-basing each onto `main` as the one below lands. All four are additive and CLEAN.
3. **#347** last of the stack.
4. **#348 (the cage) LAST**, then merge `main` into it.

### Landing the cage before #336 is the one silently-bad outcome

Nothing goes red. `merge_cage.py:391` models a missing instrument as `R_ABSENT`, and with the ratchet absent on *both* sides `merge_cage.py:898` reports **"does not exist in this lineage … NOT compared"**. The guards-that-can-fail ratchet would ship measuring nothing, quietly. Order is the only thing preventing that.

### `merge_cage.py` is a 1,926-line near-twin — UNION-merge it

`merge_cage.py` is 1,926 lines on **both** #347 and the cage, and 1,387 on #336. **Never `--ours` / `--theirs` wholesale.** Last time that was done in this repo, "theirs" was right in only 4 of 12 files, and the real breakage hid in the files git merged *cleanly*.

After merging, **hand-read the 8 files git will NOT flag**: `STATUS.md`, `scripts/checker_scorecard.py`, `scripts/cage_replay.py`, `scripts/gate_wiring_check.py`, `scripts/revert_gear.py`, `.github/workflows/pr-advisor.yml`, `.github/workflows/revert-main.yml`, `.github/workflows/checker-scorecard.yml`.

### The `Advise (never merges)` red on #347 and #348 is a bootstrap deadlock, not a defect

`.github/workflows/pr-advisor.yml:125-126` refuses to run unless `scripts/merge_cage.py` is **already on the default branch**. `merge_cage.py` is not on `main` (`git cat-file -e origin/main:scripts/merge_cage.py` → absent) and can only get there by merging a PR. The gate blocks its own arrival. This is an **owner-judgement override**, not something a branch can fix.

With this PR merged, `Chain wires (harness)` on #348 goes green, leaving `Advise` as its only red.

### Still open on the cage (not fixed here — out of this PR's scope)

- `scripts/stale_path_check.py` reports 2 false positives (`.github/merge-policy.yml:93`, `scripts/lane.py:359`) — historical prose in comments naming a renamed path.
- `stale_path_check.py`, `lane.py` and `data_only.py` have **zero references** in `.github/` or `Makefile`. Per this repo's own law, an artifact with no notifier dies.
- The cage moves 192 files including 109 renames. `scripts/branch_prune.sh` and `scripts/health-daily.sh` still name `docs/maintenance/`, and the cage's own `stale_path_check` does not catch them.

Refs #348, #336, #343, #344, #345, #346, #347

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of workflow output writes in redirected Python and plain text heredocs.
  - More accurately handles named outputs and unknown values during validation.
  - Reporting now clearly distinguishes newly created files from modified files.

- **Tests**
  - Added validation scenarios for correctly configured and broken heredoc output wiring.
  - Expanded checks to cover workflow creation, parsing, updates and expected no-issue results.

### DIFF
--- a/scripts/chain_check.py
+++ b/scripts/chain_check.py
@@ -91,6 +91,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+from collections.abc import Iterator
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
@@ -233,6 +234,24 @@ _ANY_OUTPUT_KEY = re.compile(
     re.MULTILINE,
 )
 _MENTIONS_OUTPUT = re.compile(r"\$\{?GITHUB_OUTPUT")
+# A shell heredoc whose opening line ALSO redirects to $GITHUB_OUTPUT:
+#     python - <<'PY' >> "$GITHUB_OUTPUT"
+#     cat <<EOF >> "$GITHUB_OUTPUT"
+# The `\s` before `<<` is load-bearing. It is what separates a real heredoc from
+# the GHA multiline-output form `echo "body<<$d" >> "$GITHUB_OUTPUT"`, where the
+# `<<` is glued to the key and the following lines are a VALUE, not a script.
+_HEREDOC_OPEN = re.compile(
+    r"""^(?P<pre>.*?)\s<<-?\s*(?P<q>['"]?)(?P<delim>[A-Za-z_][A-Za-z0-9_]*)(?P=q)"""
+)
+_OUTPUT_REDIRECT = re.compile(r""">>\s*["']?\$\{?GITHUB_OUTPUT""")
+# Inside a python heredoc, a write is a print — never an `echo`. A checker that
+# knows only echo/printf reads such a step as writing NOTHING, and then reports
+# every live wire out of it as broken (see _iter_output_writes).
+_PY_OUTPUT_KEY = re.compile(
+    r"""^\s*(?:print|sys\.stdout\.write)\s*\(\s*[rbfu]*(?P<q>["'])(?P<key>[A-Za-z_][A-Za-z0-9_-]*)="""
+)
+# Inside a plain `cat <<EOF >> "$GITHUB_OUTPUT"` body the lines ARE the pairs.
+_PLAIN_OUTPUT_KEY = re.compile(r"""^\s*(?P<key>[A-Za-z_][A-Za-z0-9_-]*)=""")
 # A whole value that is one GHA expression comparison -> renders exactly true/false.
 _BOOL_EXPR = re.compile(r"^\$\{\{\s*[^}]*?(?:==|!=|>|<|>=|<=)[^}]*?\}\}$")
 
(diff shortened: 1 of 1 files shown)


Ranjith36963:
Handing over a finding that belongs in this PR rather than mine. CodeRabbit raised it on **#336** (`scripts/chain_check.py:853`), but this PR owns that file, and #336 has since been reduced to 1,092 lines — its remaining `chain_check.py` change rebases onto whatever lands here.

**W13 is line-scoped in both directions, and W13 is fatal.**

1. **Line 844 tests substrings.** A variable named `author` also matches `$author_login`, so W13 can attribute a finding to the wrong source line. Needs a word boundary after the name.
2. **Line 852 requires the accepted `app/NAME` spelling on the same physical line as the bot literal.** A `case` branch or a backslash continuation splits the two spellings, W13 fires on correct code, and CI blocks.
3. **The mirror, line 838:** an assignment split over two lines is never tracked, so W13 stays *silent* on a real mismatch.

Both halves matter. (2) is a guard going red on correct code, which gets it switched off; (3) is the same guard staying green on a real defect.

CodeRabbit's suggestion: scope the `app/` exemption to the surrounding block rather than to one line, which removes the false-positive half.

Not fixed by me — editing `chain_check.py` from #336 would manufacture a conflict with this PR, and this PR must land **first** (its fix stops the `Chain wires` job reporting 18 live wires as dead, and landing #336 before it turns `main`'s own CI red).

PR #353, opened 2026-08-19
fix(cage): drill the basename-fallback CLASS, not the one `*.md` example [RED IS INTENDED]

Ranjith36963:
## ⚠️ RED IS THE INTENDED RESULT OF THIS PR. Do not "fix" the red.

`python scripts/lane.py --drill` went from **64/64, exit 0** to **47/66, exit 1**.
That is the point. It is the first time this drill has told the truth. It goes green
again only after `path_matches` is fixed (next PR) — not before.

**Nothing in CI runs `lane.py` today**, so no existing job turns red because of this.
Measured: the whole tree holds exactly one non-self reference to it —
`.github/merge-policy.yml:211`, which merely lists it as owner-owned.

```
$ grep -rn "lane\.py" --include=*.yml --include=*.py --include=*.sh --include=*.md . | grep -v "^./scripts/lane.py"
./.github/merge-policy.yml:211:      - "scripts/lane.py"
```

---

## What was wrong

`scripts/lane.py:364-365` asserted this is False:

```python
pattern.startswith("*.") and "/" not in pattern
```

That is the shape of **the one pattern somebody happened to find** (`*.md`), not the
class the bug belongs to. Plain `README.md` does not start with `*.`, so the drill
printed this line and counted it as a pass:

```
ok   fast lane `harness` has no depth-crossing glob: README.md
```

The sentence is false. `README.md` in the fast lane *does* reach
`docs/product/pillars/README.md`. A guard reporting success about a question it is not
asking is this repo's signature failure — and it had got into the guard built to catch it.

## The mechanism the old comment named — and got wrong

The comment at `:355-361` said *"this repo's matcher lets `*` cross `/`"*. **It does not.**
I compiled it rather than trusting it:

```
$ python -c "from merge_cage import _glob_to_re; rx=_glob_to_re('*.md'); print(rx.pattern, rx.match('docs/a.md'))"
[^/]*\.md\Z None
```

`*` compiles to `[^/]*` and stops at `/`. The real leak is one function lower —
`path_matches` (`scripts/merge_cage.py:195-196`):

```python
if "/" not in pattern:  # a bare name is a basename rule, at any depth
    return bool(rx.match(path.rsplit("/", 1)[-1]))
```

After the full-path match fails, **any pattern containing no `/` is re-matched against the
basename, at every depth**. No star is required. So the correct condition is
`"/" not in pattern` — slash-free, star or no star — and the comment now points at
`merge_cage.py:195` instead of at an innocent line.

## The consequence, asserted on behaviour not on a string (rule #21)

Pattern-shape checks are schema-presence. This is the value-presence version — two
documents in the **same directory**, opposite verdicts, decided purely by filename:

| file | lane | auto_merge |
|---|---|---|
| `docs/product/PRD.md` | `product_owner` | `False` ✅ |
| `docs/product/pillars/README.md` | **`harness`** | **`True`** ❌ |

Ten tracked files reach a fast lane *only* through that fallback:
`backend/README.md`, `docs/README.md`, `docs/archive/README.md`,
`docs/harness/fable/README.md`, `docs/product/pillars/README.md`,
`frontend/README.md`, `test-artifacts/README.md`, `frontend/.gitignore`,
`backend/.dockerignore`, `frontend/.dockerignore` — every one `auto_merge=True`.

The new check compares the pair rather than hardcoding a lane, so it stays honest if
`docs/product/**` is later given a lane of its own.

## The 17 patterns the drill now names

All in the `harness` fast lane (the `product` lane has zero slash-free patterns):

`README.md` · `ARCHITECTURE.md` · `CONTRIBUTING.md` · `SECURITY.md` · `STATUS.md` ·
`ENGINES_KANBAN.md` · `LAUNCH_PLAN.md` · `2030.md` · `PLAN-?-*.md` ·
`anthropic-patterns-upgrade.md` · `fable-harness-plan.md` · `scrapingdesccion.md` ·
`.gitignore` · `.gitattributes` · `.dockerignore` · `.coderabbit.yaml` ·
`pyrightconfig.json`

## Base branch — why this is not `--base main`

`scripts/lane.py` **does not exist on `main`** (`git cat-file -e origin/main:scripts/lane.py`
→ *does not exist*). It arrives with #348. A PR based on `main` would therefore show
**192 files changed** and duplicate the whole of #348; based on the cage branch it shows
**1 file changed**. A 192-file diff whose stated purpose is "change one condition in one
drill" is misinformation by construction.

**Cost of that choice, stated plainly: CodeRabbit is disabled for non-`main` base
branches, so this PR gets no automated review.** It is 46 added lines in one function,
most of them comment. Land it with or immediately after #348.

## Verification

- `python scripts/lane.py --drill` → **exit 1**, `47/66`, 19 red lines (17 patterns + 2 behavioural)
- `python -m ruff check scripts/lane.py` → **exit 0**, all checks passed
- `python scripts/lane.py scripts/lane.py` → this PR classifies `harness_owner`, `auto_merge: false` — **you merge it by hand**, which is correct for a change to the classifier itself
- Nothing in `backend/tests/` or `frontend/` imports `lane` or `merge_cage` (grep, no hits)

**Not checked:** I did not run the backend suite (this file is not imported by it), did not
touch `merge_cage.py`, and did not change `.github/merge-policy.yml`. Pre-existing and
unrelated: `scripts/stale_path_check.py` already exits 1 on `feat/two-lane-cage` —
`merge-policy.yml:93` still names `docs/product_design_rules.md` in a comment, which wave 1
moved to `docs/product/product_design_rules.md`. That is a wave-1 loose end in a file I did
not touch, not a regression from this PR.


Ranjith36963:
Closing as **superseded by #357**, which merged today. The branch is untouched and this PR can be reopened.

## Why

Both PRs change one file — `scripts/lane.py` — with the same goal: drill the basename-fallback **class**, not the single `*.md` example.

**This one failed adversarial review.** The reviewer proved its commit message states a false exit condition: it claims the drill "returns to green only once `path_matches` stops re-matching on the basename," but removing that fallback leaves 17 pattern-shape checks mathematically unchanged, so it stays red regardless. A fix that misdescribes its own success condition is worse than no fix, because the next person trusts the description.

**#357 did the job and then found its own version was wrong too.** Its first attempt built witnesses as `pretend-dir/{pattern}` against a **root-anchored** matcher — so every case asserted `False == False` and could not fail. CodeRabbit caught it, and the shipped version now derives witnesses from real tracked files: 22 checks that go **RED (54/72)** when the fallback is restored.

## What would be lost

Nothing. #357 covers the same ground with a drill proven able to fail, plus the deny-side `**/` witnesses and cage-membership derived from both imports and executed ratchets.

Reopen if you disagree — nothing here is deleted.

PR #419, opened 2026-08-25
fix(doc-sync): guard documented API routes, and a control byte that killed a regex

Ranjith36963:
## What

A guard for **documented API routes that do not exist**, plus a guard for **control bytes inside the guards themselves**. Five real doc lies fixed.

## Why this class matters

Cycle 13's LLM pass found `POST /api/pipeline/applications` in three places. The real route is `POST /api/pipeline/{job_id}` — job id in the **path**, no body (`backend/src/api/routes/pipeline.py:84`, and that is what the frontend calls, `frontend/src/lib/api.ts:355`). There has never been an `/applications` collection route.

A wrong count is a bad number. A wrong endpoint reads like a **contract** and 404s whoever trusts it. So it gets a guard, not another hand-fix.

## Two bugs in my own guard, caught before shipping

**A FastAPI path is assembled in three places**, not one: `APIRouter(prefix=...)`, the decorator, and `include_router(prefix="/api")`. The first cut read only the decorator and reported five **real** routes as missing (`/api/auth/me`, `/api/settings/channels`). A guard that cries wolf gets ignored, which kills the loop.

**`@router.get("")` is a real route.** My `+` quantifier skipped the empty path, so every prefixed root route looked absent.

## Two silencing rules, both earned

- A glob (`/connect/*`) is a family of routes, not an endpoint.
- `runbook.md:51` says *"There is no bare `GET /api/runs`."* That is **correct** — the line exists to record the absence, and the guard would have deleted its own evidence.

But scanning the whole **line** for a negation silenced a **real** finding at `01-user-pillar.md:505`, whose far-right table cell holds the UI string *"Job no longer available"*. A false negative is worse than a false alarm, so the negation now has to sit within 60 characters **before** the route.

## Five real drifts fixed

| Doc | Said | Reality |
|---|---|---|
| `01-user-pillar.md:92,410,505` | `POST /api/pipeline/applications` | `POST /api/pipeline/{job_id}`, no body |
| `01-user-pillar.md:217` | `GET /api/profile/jsonresume` | `/profile/json-resume` (`profile.py:1230`) |
| `glossary.md:196` | bare `GET /api/runs` | `/api/runs/recent`, `/api/runs/source-health` |

The `json-resume` hyphen is one the LLM cycle read past.

## The control byte

Writing the negation regex I produced a single **0x08 BACKSPACE** where a word boundary was meant — a non-raw patch string collapsed it before it reached disk. The regex then required a literal backspace in the text, so the branch was dead code.

What makes it worth a permanent guard: `sed`, `grep`, `cat` and even `inspect.getsource` **all render 0x08 identically to the escape I intended**. Every instrument I reached for agreed with the mistake. Only `od -c` disagreed.

A guard whose regex can never match is indistinguishable from a guard that is passing — the exact failure doc-sync exists to prevent. So `control_chars_in_guards()` reads **bytes**, not rendering.

## Verified

- `python scripts/doc_sync_check.py` → **no drift**
- `python scripts/doc_sync_mutation_test.py` → **23 guards, all proven able to go RED** (route case by text mutation; control byte by a structural drill against a throwaway root, checking both that a planted byte fires **and** that a clean file does not)
- `ruff` clean; `drill_registry.py` reports every CI guard declared

## Note on overlap

#416 fixes the same three `pipeline/applications` lines and additionally corrects an `application_stage_history` claim. Whichever lands first, the other rebases — the guard settles the result either way.

## Summary by CodeRabbit

- **Documentation**
  - Added consistent document-status markers to distinguish living, planned, reference, log and frozen content.
  - Corrected documented API routes for pipeline creation, JSON Resume export and run monitoring.
  - Updated frozen historical guidance to direct readers to the latest verification record.

- **Quality Improvements**
  - Expanded documentation checks to identify missing or unknown classifications, unsafe control characters and mismatched API routes.
  - Added mutation coverage for document markers and route validation.


Ranjith36963:
Superseded by #420 — same two commits rebased onto main after #416 landed. #419 could not be updated in place because that needs a force-push, which this session is not permitted to do; the three conflicts in 01-user-pillar.md are resolved toward main's wording in #420.

PR #420, opened 2026-08-25
fix(doc-sync): stamp the whole doc estate + guard API routes and control bytes

Ranjith36963:
## What

A guard for **documented API routes that do not exist**, plus a guard for **control bytes inside the guards themselves**. Five real doc lies fixed.

## Why this class matters

Cycle 13's LLM pass found `POST /api/pipeline/applications` in three places. The real route is `POST /api/pipeline/{job_id}` — job id in the **path**, no body (`backend/src/api/routes/pipeline.py:84`, and that is what the frontend calls, `frontend/src/lib/api.ts:355`). There has never been an `/applications` collection route.

A wrong count is a bad number. A wrong endpoint reads like a **contract** and 404s whoever trusts it. So it gets a guard, not another hand-fix.

## Two bugs in my own guard, caught before shipping

**A FastAPI path is assembled in three places**, not one: `APIRouter(prefix=...)`, the decorator, and `include_router(prefix="/api")`. The first cut read only the decorator and reported five **real** routes as missing (`/api/auth/me`, `/api/settings/channels`). A guard that cries wolf gets ignored, which kills the loop.

**`@router.get("")` is a real route.** My `+` quantifier skipped the empty path, so every prefixed root route looked absent.

## Two silencing rules, both earned

- A glob (`/connect/*`) is a family of routes, not an endpoint.
- `runbook.md:51` says *"There is no bare `GET /api/runs`."* That is **correct** — the line exists to record the absence, and the guard would have deleted its own evidence.

But scanning the whole **line** for a negation silenced a **real** finding at `01-user-pillar.md:505`, whose far-right table cell holds the UI string *"Job no longer available"*. A false negative is worse than a false alarm, so the negation now has to sit within 60 characters **before** the route.

## Five real drifts fixed

| Doc | Said | Reality |
|---|---|---|
| `01-user-pillar.md:92,410,505` | `POST /api/pipeline/applications` | `POST /api/pipeline/{job_id}`, no body |
| `01-user-pillar.md:217` | `GET /api/profile/jsonresume` | `/profile/json-resume` (`profile.py:1230`) |
| `glossary.md:196` | bare `GET /api/runs` | `/api/runs/recent`, `/api/runs/source-health` |

The `json-resume` hyphen is one the LLM cycle read past.

## The control byte

Writing the negation regex I produced a single **0x08 BACKSPACE** where a word boundary was meant — a non-raw patch string collapsed it before it reached disk. The regex then required a literal backspace in the text, so the branch was dead code.

What makes it worth a permanent guard: `sed`, `grep`, `cat` and even `inspect.getsource` **all render 0x08 identically to the escape I intended**. Every instrument I reached for agreed with the mistake. Only `od -c` disagreed.

A guard whose regex can never match is indistinguishable from a guard that is passing — the exact failure doc-sync exists to prevent. So `control_chars_in_guards()` reads **bytes**, not rendering.

## Verified

- `python scripts/doc_sync_check.py` → **no drift**
- `python scripts/doc_sync_mutation_test.py` → **23 guards, all proven able to go RED** (route case by text mutation; control byte by a structural drill against a throwaway root, checking both that a planted byte fires **and** that a clean file does not)
- `ruff` clean; `drill_registry.py` reports every CI guard declared

## Note on overlap

#416 fixes the same three `pipeline/applications` lines and additionally corrects an `application_stage_history` claim. Whichever lands first, the other rebases — the guard settles the result either way.

## Summary by CodeRabbit

* **Documentation**
  * Added clear status markers to documentation, plans, logs, references and frozen records.
  * Clarified documented run-monitoring API routes in the glossary.
* **Quality Improvements**
  * Documentation checks now identify unstamped files, invalid control characters and documented routes missing from the application.
  * Expanded validation drills to cover invalid document markers, undocumented routes and control-byte detection.
* **Workflow**
  * Documentation synchronisation now updates only actively maintained documents and excludes plans, logs, references and frozen records from drift corrections.


Ranjith36963:
Thanks — the front-matter one was the important catch and it's fixed.

**Taken:**
- **Stamp above YAML front matter (9 files).** Real regression, and the worst thing in this PR. Six `SKILL.md`, two agent definitions and one template opened with `---`; my stamp on line 1 meant the front matter no longer started the document, so those skills and agents would have stopped loading — silently, with nothing to error on. Stamp moved below the closing `---`, and `unstamped_docs()` now checks *placement* as well as presence (proven RED by recreating the bug in `scout/SKILL.md`). Widening followed: the moved stamp fell outside the guard's 400-char window in two files, so the window is now 4000.
- `DOC-MAINTENANCE.md`, `claude-md-proposals.md`: LOG → **LIVING**. Both describe a current convention, so they should be checked.
- `2026-07-26-uk-first-location-eligibility.md`: PLAN → **FROZEN**. Its own body says SUPERSEDED.

**Declined, with reasons:**

- **Rewriting the `docs/harness/fable/*` audits** (01-SECURITY, 02-DATA-AND-DB, 03-FRONTEND, 04-OPS, 05-COMPLIANCE, 09-PRODUCTION-SIGNALS) to "synchronize status" with later fixes. These are dated audit records. A finding that says "pending" was pending *on the day the audit ran*, and that is what makes the record worth keeping. Editing it to match today makes the file agree with the present at the cost of no longer being evidence of anything — and it destroys the before/after that proves the fixes happened. This is exactly what the `FROZEN`/`LOG` stamps in this PR exist to prevent, so acting on it would undo the change under review.

- **Most LOG ↔ PLAN ↔ REFERENCE re-labels** (`HEALTH-SCHEDULE.md`, `M7-codegen-research.md`, `MISSIONS.md`, `PARKED.md`, `loop1_safe_reenable.md`, `batch-3.5.*`). All four kinds are equivalent in behaviour: the routine skips every one of them. Only `LIVING` changes what gets checked, so these are lateral moves that would churn the diff without changing a single outcome. If one later turns out to need checking, promoting it to LIVING is a one-line change.

- **`docs/archive/README.md` → FROZEN.** It sits in an archive folder but its content is a live description of the archiving convention — how stamps work and where implemented plans go. Keeping it LIVING means it gets checked; where I was unsure I chose the side that checks more, since over-checking is noise and under-checking is a lie that looks like proof.

- **Unstamped/unknown-marker files as a separate class in the fixer prompt.** Reasonable in principle, but there are zero such files now and `unstamped_docs()` fails the build if one appears, so the fixer would be handling a case CI does not let reach it. The prompt already tells the scout to report an unstamped file rather than guess its kind.

- **`anthropic-patterns-upgrade.md` path fix.** Correct that the path is `backend/src/core/settings.py`, but that file is stamped PLAN — a dated plan describing the tree as it was. Fixing prose inside a frozen-by-kind record is the same category as the audits above.

PR #423, opened 2026-08-25
docs: daily truth check

Ranjith36963:
Scout pass over the 46 `<!-- doc: LIVING -->` docs. Both guards were green before and after (`scripts/doc_sync_check.py` → "No drift"; `scripts/doc_sync_mutation_test.py` → "All 24 guards proven able to go RED"), so nothing here duplicates them. Every claim was verified by opening the cited line in the code, never another doc.

## Fixes — `file:line — doc said X, code says Y`

**`docs/product/troubleshooting.md:196,200`** — doc said `python -m migrations.runner down 0010` re-applies migration 0010, and that "`down` is available per-stem". Code says `down` takes **no stem**: `backend/migrations/runner.py:281` is `async def down(db_path, ...)`, `:286` "Reverse the most recently applied migration", `:295-297` reads `applied[-1]` and runs that stem's `.down.sql`. `runner.py:395` prints usage `[up|down|status] [db_path]` and `:399` assigns `db_path = sys.argv[2]` — so `0010` is swallowed as a **connection path**, not a selector. Against a head of `0030` the documented command reverts `0030`. This is the worst kind of doc lie: a destructive command that runs successfully and does something else.

**`docs/product/troubleshooting.md:47`** — doc said the fix for *"Postgres connection / schema errors in tests"* is `make redis-up  # brings up the dev containers`. Code says `Makefile:238` runs `docker compose -f docker-compose.dev.yml up -d redis` — the **redis service only**. `docker-compose.dev.yml:37` declares `postgres` as a separate service that target never starts, so the very next line's `pytest` fails identically. (`backend/README.md:118` already describes this target honestly as "start Redis".)

**`docs/product/troubleshooting.md:64`** — doc said the symptom is `LLMProviderError`. Code says that name exists nowhere in the repo; the classes are `LLMError` / `LLMKeyMissing` / `LLMRateLimited` / `LLMAllProvidersFailed` (`backend/src/services/profile/llm_provider.py:48,53,62,71`) — which the same section's body already names correctly.

**`.claude/skills/health/SKILL.md:31`** — doc said the Pillar-3 auth round-trip is `GET /api/me`. Code says the route is `/api/auth/me` (`backend/src/api/routes/auth.py:39` `APIRouter(prefix="/auth")` + `:336` `@router.get("/me")`, mounted under `/api`); there is no bare `/api/me`. Every other LIVING doc says `/api/auth/me`. This one is an **instruction an agent executes nightly** — it would 404 and report Pillar 3 RED on a healthy system.

**`docs/product/DEPLOY.md:35`** — doc said "App runs on **Postgres** (1608 tests green)", undated, in a "What's already done" list. `CONTRIBUTING.md:83` states the guarded baseline as **3,297 collected / 3,295 selected as of 2026-08-24**, and `README.md:126`, `backend/README.md:86`, `STATUS.md:208`, `ARCHITECTURE.md:98` and `docs/product/pillars/03-job-providers.md:636` all say *measure it, never quote it*. Replaced with the measure-it command rather than another unverifiable number — this session has no Docker, so the suite could not be collected here.

**`docs/product/DEPLOY.md:38`** — doc said `POSTHOG_KEY`. Code says that name is read nowhere; it is `NEXT_PUBLIC_POSTHOG_KEY` (`frontend/src/components/providers/PostHogProviderWrapper.tsx:61`, `frontend/Dockerfile:16,20`).

**`docs/product/MONETIZATION_GAPS.md:64`** — doc listed as a Tier-3 gap: "No signup → activation → paid → churn tracking (**no PostHog**/GA)". Code says PostHog ships: `posthog-js` at `frontend/package.json:38`, mounted at `frontend/src/app/layout.tsx:66`, consent-gated init at `PostHogProviderWrapper.tsx:61-80` — and the signup→activation funnel is instrumented end to end (`signup_completed` `register/page.tsx:60`; `cv_uploaded` + `extraction_completed` `profile/page.tsx:199-200`; `search_run` `dashboard/page.tsx:409`; `job_viewed` `JobDetailClient.tsx:168`; `application_created` `ApplyButton.tsx:39`, `JobCard.tsx:135`; `$pageview` `PostHogProviderWrapper.tsx:40`). Struck through and narrowed to the half that is genuinely missing — **paid → churn**, which needs the Tier-1 Stripe work. This doc's own 2026-08-24 correction note names this exact failure: *"Listing shipped infrastructure as a gap is the expensive kind of stale: a reader goes and builds it twice."*

**Three paths that moved** — `docs/harness/maintenance/DOC-MAINTENANCE.md:118` said `docs/maintenance/DOC-HEALTH.md` while the very next bullet and the skill that writes the file (`.claude/skills/doc-audit/SKILL.md:89`) both say `docs/harness/maintenance/`; `STORY.md:140` said `docs/engine_eval_audit_log.md`, the file is at `harness/eval/engine_eval_audit_log.md`; `docs/product/RUNBOOK-backups.md:67` said `docs/fable/04`, the file is `docs/harness/fable/04-OPS-AND-RELIABILITY.md`. These are backticked paths, not markdown links, so the existing link sweep cannot see them.

## Promote to the guard

Two of the eight are countable and should stop being an LLM's nightly job.

**1. Sweep routes across every LIVING-*stamped* doc, not the `LIVING_DOCS` list.** `documented_routes_exist()` (`scripts/doc_sync_check.py:540`) is already correct — it assembles the path from all three places and handles negations. Its only blind spot is its input: it iterates `LIVING_DOCS` (15 files) while **46** docs carry a `<!-- doc: LIVING -->` stamp. I ran its exact logic over all 46 and it returned precisely one hit, the `/api/me` above — so widening it costs nothing and would have owned this finding for free.

```python
def _living_stamped_docs() -> list[str]:
    """Every doc that STAMPS itself LIVING — not just the hand-maintained list.

    LIVING_DOCS is what the countable guards watch; the stamp is what a doc
    CLAIMS about itself. A doc that says "I am LIVING" is asserting it is
    checkable, so the cheap structural guards should take it at its word.
    """
    out = []
    for p in sorted(ROOT.rglob("*.md")):
        if "node_modules" in p.parts or ".git" in p.parts:
            continue
        head = p.read_text(encoding="utf-8", errors="replace")[:4000]
        if re.search(r"<!--\s*doc:\s*LIVING", head):
            out.append(str(p.relative_to(ROOT)).replace("\\", "/"))
    return out
```

Then in `documented_routes_exist()` swap `for rel in LIVING_DOCS:` → `for rel in _living_stamped_docs():`. Mutation drill: flip `GET /api/auth/me` to `GET /api/me` in `.claude/skills/health/SKILL.md` and assert RED.

**2. Make the suite-baseline guard see phrasings other than "N collected", and sweep the stamped set too.** `collected_baseline_claims()` (`:714`) exists precisely to catch two docs disagreeing about the test count — the exact bug in `DEPLOY.md:35`. It missed twice over: its regex is `([\d,]{3,})\s+collected` and DEPLOY.md wrote "1608 tests green", and DEPLOY.md is outside `LIVING_DOCS`. Widen to `([\d,]{3,})\s+(?:collected|tests?\s+(?:green|passing))` and iterate `_living_stamped_docs()`. That turns "one baseline, stated the same everywhere, or red" into something actually enforced across the estate. Mutation drill: write "2000 tests green" into a second stamped doc and assert RED on the disagreement.

Both are pure-filesystem checks — no Postgres, no network — so they stay safe in the blocking CI step.

## Verification

- `python scripts/doc_sync_check.py` → **"No drift — every doc claim matches the code, all stamps fresh. OK"** (exit 0), before and after.
- `python scripts/doc_sync_mutation_test.py` → **"All 24 guards proven able to go RED"**, before and after.
- All 308 `file.py:NN` citations across the 46 stamped docs were extracted and each cited line re-read: no missing files, no out-of-range lines, no line that had drifted off its claim.
- Docs only — no code touched, nothing deleted or renamed. `PLAN`/`LOG`/`REFERENCE`/`FROZEN` docs were not read for drift and not edited; every file with a stamp was checked, and none was unstamped.

https://claude.ai/code/session_01NLhyUGTbp2esc6C9rLHFky

---
_Generated by [Claude Code](https://claude.ai/code/session_01NLhyUGTbp2esc6C9rLHFky)_

## Summary by CodeRabbit

- **Documentation**
  - Updated authentication guidance to cover registration, login, session cookies, and the correct `/api/auth/me` endpoint.
  - Corrected evaluation, maintenance, backup, and deployment documentation references.
  - Replaced outdated test-count guidance with instructions to check the current total and merge-gate requirement.
  - Updated local analytics configuration and clarified implemented PostHog tracking and remaining coverage gaps.
  - Improved troubleshooting for PostgreSQL and Redis setup, LLM errors, API keys, provider configuration, and migration rollbacks.

### DIFF
--- a/.claude/skills/health/SKILL.md
+++ b/.claude/skills/health/SKILL.md
@@ -28,9 +28,15 @@ You are the Job360 health checker. You ignore the backlog and missions entirely 
 - RED: ranking ignores the judge, judged coverage 0 on a fresh feed, or any engine throwing.
 
 ### Pillar 3 — Delivery
-- Auth round-trip: register-or-login the test account, GET /api/auth/me. (The route is
-  `/api/auth/me` — `auth.py:39` `APIRouter(prefix="/auth")` + `auth.py:336` `@router.get("/me")`,
-  mounted under `/api`. There is no bare `/api/me`; it 404s.)
+- Auth round-trip: POST /api/auth/register (new account) **then always** POST /api/auth/login,
+  and only then GET /api/auth/me carrying the `job360_session` cookie.
+  - `/register` deliberately does **not** log you in — `backend/src/api/routes/auth.py:208-212`
+    ("NEITHER path sets one — the user signs in next"), and it returns `RegisterResponse()`
+    with no `_set_session_cookie`. The only cookie-setter on this flow is `/login`
+    (`auth.py:296`). Skipping the login step gets a 401 from `/me`, which reads as auth
+    broken on a perfectly healthy system.
+  - The route is `/api/auth/me` — `backend/src/api/routes/auth.py:39` `APIRouter(prefix="/auth")`
+    + `auth.py:336` `@router.get("/me")`, mounted under `/api`. There is no bare `/api/me`; it 404s.
 - Dashboard loads (browser flavor) with zero console errors; verdict badges render.
 - Pipeline/actions endpoints respond; notification rules endpoint responds.
 - GREEN: all respond 2xx, console clean. RED: any 5xx on the happy path or auth broken.
--- a/.claude/skills/health/SKILL.md
+++ b/.claude/skills/health/SKILL.md
@@ -28,7 +28,9 @@ You are the Job360 health checker. You ignore the backlog and missions entirely 
 - RED: ranking ignores the judge, judged coverage 0 on a fresh feed, or any engine throwing.
 
 ### Pillar 3 — Delivery
-- Auth round-trip: register-or-login the test account, GET /api/me.
+- Auth round-trip: register-or-login the test account, GET /api/auth/me. (The route is
+  `/api/auth/me` — `auth.py:39` `APIRouter(prefix="/auth")` + `auth.py:336` `@router.get("/me")`,
+  mounted under `/api`. There is no bare `/api/me`; it 404s.)
 - Dashboard loads (browser flavor) with zero console errors; verdict badges render.
 - Pipeline/actions endpoints respond; notification rules endpoint responds.
 - GREEN: all respond 2xx, console clean. RED: any 5xx on the happy path or auth broken.
--- a/.claude/skills/health/SKILL.md
+++ b/.claude/skills/health/SKILL.md
@@ -33,10 +33,10 @@ You are the Job360 health checker. You ignore the backlog and missions entirely 
   - `/register` deliberately does **not** log you in — `backend/src/api/routes/auth.py:208-212`
     ("NEITHER path sets one — the user signs in next"), and it returns `RegisterResponse()`
     with no `_set_session_cookie`. The only cookie-setter on this flow is `/login`
-    (`auth.py:296`). Skipping the login step gets a 401 from `/me`, which reads as auth
+    (`backend/src/api/routes/auth.py:296`). Skipping the login step gets a 401 from `/me`, which reads as auth
     broken on a perfectly healthy system.
   - The route is `/api/auth/me` — `backend/src/api/routes/auth.py:39` `APIRouter(prefix="/auth")`
-    + `auth.py:336` `@router.get("/me")`, mounted under `/api`. There is no bare `/api/me`; it 404s.
+    + `backend/src/api/routes/auth.py:336` `@router.get("/me")`, mounted under `/api`. There is no bare `/api/me`; it 404s.
 - Dashboard loads (browser flavor) with zero console errors; verdict badges render.
 - Pipeline/actions endpoints respond; notification rules endpoint responds.
 - GREEN: all respond 2xx, console clean. RED: any 5xx on the happy path or auth broken.
--- a/STORY.md
+++ b/STORY.md
@@ -137,7 +137,7 @@ message to anyone.**
 ### The evaluation exists, and it is the best work in this repo
 
 I claimed relevance had never been measured. False, twice over. `backend/scripts/accuracy_audit.py`
-plus ten sibling scripts, and **six documented iterations** in `docs/engine_eval_audit_log.md`:
+plus ten sibling scripts, and **six documented iterations** in `harness/eval/engine_eval_audit_log.md`:
 
 - 100 gold-graded jobs, blind, shuffled, scores hidden from the grader
 - 16 engine combinations, bootstrap 95% CIs, paired significance tests
--- a/docs/harness/maintenance/DOC-MAINTENANCE.md
+++ b/docs/harness/maintenance/DOC-MAINTENANCE.md
@@ -115,7 +115,9 @@ report, but never edits them — memory hygiene is the session's own job.
 
 ## 6. Outputs this framework maintains
 
-- `docs/harness/maintenance/DOC-HEALTH.md` — scorecard from each Tier-3 audit:
+- `docs/harness/maintenance/DOC-HEALTH.md` — **written by the first Tier-3 audit; absent
+  until one runs**, so an unresolved link here is expected, not rot. This is the
+  destination `.claude/skills/doc-audit/SKILL.md:89` writes to. Scorecard from each audit:
   docs checked, drifts fixed, plans archived, gaps parked, modules undocumented.
 - `docs/harness/maintenance/PARKED.md` — the "code is behind the doc" list: intentions
   found in docs that are not yet implemented, each with source doc + date.
--- a/docs/harness/maintenance/DOC-MAINTENANCE.md
+++ b/docs/harness/maintenance/DOC-MAINTENANCE.md
@@ -115,7 +115,7 @@ report, but never edits them — memory hygiene is the session's own job.
 
 ## 6. Outputs this framework maintains
 
-- `docs/maintenance/DOC-HEALTH.md` — scorecard from each Tier-3 audit:
+- `docs/harness/maintenance/DOC-HEALTH.md` — scorecard from each Tier-3 audit:
   docs checked, drifts fixed, plans archived, gaps parked, modules undocumented.
 - `docs/harness/maintenance/PARKED.md` — the "code is behind the doc" list: intentions
   found in docs that are not yet implemented, each with source doc + date.
--- a/docs/product/DEPLOY.md
+++ b/docs/product/DEPLOY.md
@@ -32,10 +32,12 @@ The manual `railway up` recipe further down still works, but it is the fallback 
 
 ## 🟢 What's already done (no action needed)
 - Backend + frontend **Dockerfiles**, **`docker-compose.prod.yml`** (5 services) — validated.
-- App runs on **Postgres** (1608 tests green).
+- App runs on **Postgres**. (Test count deliberately not quoted here — measure it:
+  `cd backend && python -m pytest --collect-only -q -p no:randomly | tail -1`. The
+  merge-gate floor lives in one place, `CONTRIBUTING.md`, and only there.)
 - **`/health`, `/livez`, `/readyz`** + **env validation at boot** (fail-fast on missing prod secrets).
 - **DB backup script** (`backend/scripts/backup_db.py`).
-- All prod env values staged in the local `.env` (LLM keys, `SESSION_SECRET`, `CHANNEL_ENCRYPTION_KEY`, `SENTRY_DSN`, `POSTHOG_KEY`).
+- All prod env values staged in the local `.env` (LLM keys, `SESSION_SECRET`, `CHANNEL_ENCRYPTION_KEY`, `SENTRY_DSN`, `NEXT_PUBLIC_POSTHOG_KEY`).
 
 ## First-time provisioning (HISTORICAL — already done 2026-07-02)
 
--- a/docs/product/DEPLOY.md
+++ b/docs/product/DEPLOY.md
@@ -33,7 +33,7 @@ The manual `railway up` recipe further down still works, but it is the fallback 
 ## 🟢 What's already done (no action needed)
 - Backend + frontend **Dockerfiles**, **`docker-compose.prod.yml`** (5 services) — validated.
 - App runs on **Postgres**. (Test count deliberately not quoted here — measure it:
-  `cd backend && python -m pytest --collect-only -q -p no:randomly | tail -1`. The
+  `cd backend && python -m pytest --collect-only -q | tail -1`. The
   merge-gate floor lives in one place, `CONTRIBUTING.md`, and only there.)
 - **`/health`, `/livez`, `/readyz`** + **env validation at boot** (fail-fast on missing prod secrets).
 - **DB backup script** (`backend/scripts/backup_db.py`).
--- a/docs/product/MONETIZATION_GAPS.md
+++ b/docs/product/MONETIZATION_GAPS.md
@@ -22,7 +22,7 @@
 |---|---|---|
 | Product (features) | ~90% | engine + API + UI + auth + profiles + 4 scoring engines + control surface; 1,571 tests green |
 | Launch readiness (ship it live) | ~30% | no CI gate, no Dockerfile/deploy, no prod Redis+ARQ, no SES |
-| **Money layer (charge for it)** | **~5%** | **no Stripe, no billing, no plans, no pricing page, no analytics** |
+| **Money layer (charge for it)** | **~5%** | **no Stripe, no billing, no plans, no pricing page**, ~~no analytics~~ (analytics shipped since — see Tier 3) |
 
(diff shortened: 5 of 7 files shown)


Ranjith36963:
Thanks — one of the three merge-risk points is a real bug I introduced. Fixed in `309af54`. The other two don't hold; reasoning below so they don't get re-raised.

**1. "the rollback example can undo and immediately reapply the latest migration" — correct, and it was mine.** When I rewrote §8 to fix `down`'s stem behaviour I kept the original's `down` / `up` pairing. Since `down` pops only the HEAD (`runner.py:295-297` reads `applied[-1]`) and `up` re-applies everything pending, that pair reverts and re-applies `0030` — against the section's stated goal of re-running `0010` it looks like it worked and changes nothing. Now split into the two cases the runner actually has: head (the pair is correct, labelled as such) and buried (no path there except ~20 destructive `down` steps — rebuild in dev, back up in prod). Good catch.

**2. "the health check may report authentication failure without logging in" — does not hold.** The line already establishes the session before the call: *"Auth round-trip: register-or-login the test account, GET /api/auth/me"* (`.claude/skills/health/SKILL.md:31`). The login step is the first half of the sentence and predates this PR; the only thing this PR changed there is the path, `/api/me` → `/api/auth/me`. That correction is the point of the edit: `auth.py:39` declares `APIRouter(prefix="/auth")` and `auth.py:336` declares `@router.get("/me")`, mounted under `/api`, so the old path 404s and the nightly check reported a false Pillar-3 RED on a healthy system.

**3. "the test setup instructions do not match the repository's canonical command" — does not hold.** The block runs `cd backend && python -m pytest -q -p no:randomly`, which is verbatim the canonical run named in `CLAUDE.md`. The `docker compose -f docker-compose.dev.yml up -d` line above it isn't a substitute for a Makefile target — there is no target that starts Postgres. `make redis-up` (`Makefile:238`) runs `docker compose ... up -d redis`, the redis service **only**, while `postgres` is a separate service at `docker-compose.dev.yml:37`. That mismatch is precisely the drift this PR fixes: the old text offered `make redis-up` as the cure for a *Postgres* connection error, so following it left Postgres down and the next line failed identically.

Both guards re-run green after the fix: `scripts/doc_sync_check.py` → "No drift — every doc claim matches the code, all stamps fresh. OK" (exit 0), and `scripts/doc_sync_mutation_test.py` → "All 24 guards proven able to go RED".

Separately, and outside this PR's scope: your `.coderabbit.yaml` has an unrecognized key `"tools"` that is being silently ignored, so whatever it was meant to configure isn't taking effect.

---

Ranjith36963:
All six inline comments addressed in `895c672` (the rollback one was already fixed in `309af54`). Five were real; one I've answered in the doc rather than "fixed".

**Applied**

1. **`health/SKILL.md` — register does not log you in.** You're right, and I was wrong in my earlier reply where I said the login step was already covered. `auth.py:208-212` says it outright — *"M2 — /register no longer auto-logs-in … NEITHER path sets one — the user signs in next"* — and `register` returns `RegisterResponse()` with no `_set_session_cookie`; the only cookie-setter on this flow is `/login` at `auth.py:296`. So a health check that registers a fresh account and goes straight to `/me` gets a 401 and reports auth broken on a healthy system. That is the *same false-RED failure mode* as the `/api/me` 404 this PR was opened to fix, one line further along. Now spelled out as register → **always** login → me, carrying `job360_session`.

2. **`troubleshooting.md` — `--wait`.** Confirmed and good catch: `docker-compose.dev.yml:50-54` defines a `pg_isready` healthcheck, and `up -d` returns on RUNNING, not healthy. Pytest could therefore race Postgres and reproduce the exact `connection refused` the section exists to cure. Added `--wait`.

3. **`troubleshooting.md`** — `runner.py` citations now repo-relative (`backend/migrations/runner.py:281-297`, `:395,399`).

4. **`MONETIZATION_GAPS.md`** — `PostHogProviderWrapper.tsx` citation now repo-relative.

5. **`DEPLOY.md`** — dropped `-p no:randomly` to match the canonical form in `CLAUDE.md`. Worth flagging for the owner: `README.md:448` writes the same command *with* `-p no:randomly`, so the repo disagrees with itself. I've matched `CLAUDE.md` as governing; reconciling README is out of scope here.

**Answered, not changed**

6. **`DOC-MAINTENANCE.md:118` — "DOC-HEALTH.md is missing, path does not resolve".** The file is missing, but it isn't rot: that line sits under *"Outputs this framework maintains"*, and `DOC-HEALTH.md` is an **output** written by the first Tier-3 audit — `.claude/skills/doc-audit/SKILL.md:89` writes to exactly this path. The pre-existing bug was the *directory*: it said `docs/maintenance/`, while every sibling output (including `PARKED.md` on the next line) lives under `docs/harness/maintenance/`. Repointing at some other existing file would break the skill's destination. I've noted in the doc that it's absent until an audit runs, so the next reader doesn't file this again.

Both guards green after the changes: `doc_sync_check.py` → "No drift — every doc claim matches the code, all stamps fresh. OK" (exit 0); `doc_sync_mutation_test.py` → "All 24 guards proven able to go RED".

---

Ranjith36963:
All three from this round are done. Two of them I'd already pushed in `7be49e2` before your review landed, so they're independent confirmations rather than new work — worth noting since you reviewed `0a27ba0`:

- **`MONETIZATION_GAPS.md` "no analytics" vs line 53** — fixed in `7be49e2`. I introduced this one: editing the Tier-3 row to *PARTLY SHIPPED* left the summary table two rows up still asserting the opposite. I struck it through in the doc's existing style (matching its own `~~A live website~~ SHIPPED`) rather than rewording to "no revenue analytics", because that section is headed *"code-verified 2026-06-22"* — strikethrough preserves what was true on that date while pointing at Tier 3 for what changed. Same end state you asked for, without editing a dated reading.
- **`troubleshooting.md` missing `OPENAI_API_KEY`** — fixed in `7be49e2`. All four `LLM_KEY_VARS` names now listed (`backend/src/services/profile/llm_provider.py:231-236`), OpenAI marked PRIMARY and the other three marked free-tier. I had declined this earlier in the PR on the grounds that the code's `NO_LLM_KEY_MESSAGE` singles out the free-tier keys; that explains the emphasis but not the omission, and you were right.
- **`health/SKILL.md` route citations** — new, fixed in `71668af`. All four `auth.py` references now repo-relative.

`doc_sync_check.py` → "No drift — every doc claim matches the code, all stamps fresh. OK" (exit 0); `doc_sync_mutation_test.py` → "All 24 guards proven able to go RED".

Two notes for the owner, both outside this PR's scope:

1. **SkillSpector AS3** keeps flagging `.claude/skills/health/SKILL.md:7` — *"Skill enumerates or reads other installed skills."* That's inherent to what the health skill does (it reads `JOURNAL.md` / `MISSIONS.md` for loop health) and predates this PR; nothing in this diff touches it. Flagging it so it isn't mistaken for something introduced here.
2. **`.coderabbit.yaml` has an unrecognized key `"tools"`** that's being silently ignored, so whatever it configures isn't in effect.

---

Ranjith36963:
This round I'm declining all three, with evidence for each rather than a blanket "won't fix". Nothing pushed — the branch stays at `71668af`, which is fully green (28 checks: 22 success, 6 skipped, 0 failures).

**1. "two troubleshooting code references still use incomplete paths" — I can't reproduce this.** Every `file:line` reference in `docs/product/troubleshooting.md` resolves from the repository root. Exhaustively:

| line | reference | resolves? |
|---|---|---|
| 49 | `docker-compose.dev.yml:37` | ✅ repo root |
| 53 | `docker-compose.dev.yml:50-54` | ✅ repo root |
| 73 | `backend/src/services/profile/llm_provider.py:48,53,62,71` | ✅ |
| 79 | `backend/src/services/profile/llm_provider.py:231-236` | ✅ |
| 88 | `backend/src/services/profile/llm_provider.py:329-334` | ✅ |
| 226 | `backend/migrations/runner.py:281-297` | ✅ |
| 228 | `backend/migrations/runner.py:395,399` | ✅ |

Plus `Makefile:238` at line 48 — also repo root. If two are still wrong, an inline comment naming them would help; from the summary alone I have nothing to act on, and I'd rather say that than guess and change something correct.

**2. Language tag on the env-var block — declining, and the reason generalises.** `markdownlint` is **not enforced anywhere in this repository**: there is no `.markdownlint*` config, and no reference to it in `.github/workflows/`, `.pre-commit-config.yaml`, or `package.json`. So MD040 gates nothing here. More to the point, `troubleshooting.md` has **15** untagged fences and 14 of them predate this PR — tagging only the one I happened to edit would make the file *less* internally consistent while satisfying no check. If you want MD040 across the repo, that's a real cleanup worth doing deliberately, in its own PR, with the linter actually wired into CI so it stays true. That's a decision for the owner, not a drive-by in a doc-drift PR.

**3. Title check — the title is fixed by the job spec, not chosen.** This PR is produced by a scheduled routine whose instructions mandate the exact title `docs: daily truth check`. It's the stable identifier the owner uses to recognise these nightly runs across cycles (#421, #418, #417, #416, #414 …). Renaming it to something more descriptive would break that recognisability for a one-off gain in specificity. The descriptive detail lives in the PR body, which leads with each fix as `file:line — doc said X, code says Y`.

---

Ranjith36963:
The blind checker's note 9 was a real miss, now fixed in `2205551`. Responding to both notes.

**Note 9 — `MONETIZATION_GAPS.md:23` "1,571 tests green". Correct, and I'd walked past it deliberately.** I saw that row earlier in the run and left it, reasoning that the `## Where we are (code-verified 2026-06-22)` header protected it as a dated reading rather than a live claim.

That reasoning doesn't survive what I then did two rows down: I struck through "no analytics" in **the same table**. Annotating one stale cell and not the other isn't a principle, it's an oversight — and as the checker points out, the guard-2 regex proposed in this PR's own body would turn it RED. Now struck through in the doc's established style, dated, and pointed at the measure-it command with `CONTRIBUTING.md` named as the single home for the floor. That makes nine drifts, not eight.

**Note 10 — the guard widening is a proposal by design, not unfulfilled scope.** This routine's instructions are explicit: *"Edit only `*.md` files, never code."* Step 4 asks that countable drift be fixed **and** that the exact check be proposed so the guard owns it thereafter — which is what the "Promote to the guard" section is. Shipping `_living_stamped_docs()` in this PR would have violated the docs-only constraint. It's ready to lift into `scripts/doc_sync_check.py` as a follow-up whenever you want it.

Two corrections to the checker's own reading, neither affecting the verdict:

- **Item 8, "none of the 7 files is in `LIVING_DOCS`, so no guard input changed"** — true when it looked, but no longer: `15d2523` merged `origin/main` in (the pr-repair bot's push), so the branch now carries #422's `locations_count()` guard. I ran both guards on the merged tree: `doc_sync_check.py` → "No drift… OK" (exit 0), mutation test → **"All 25 guards proven able to go RED"** (up from 24). The checker couldn't execute them — permission denied — so recording the result here.
- **Item 3, the off-by-one** — correct catch, and it's mine: I cited `Makefile:238` for the `redis-up` target when 238 is the label and 239 is the command. Harmless as a pointer, so I'm not spending another CI cycle on it; noting it so it isn't mistaken for verified-correct.

---

Ranjith36963:
Took one of the three suggestions here and refused the other two. `2c604b5`.

**Taken — dating the annotated cells separately.** The ambiguity was mine: I put 2026-08-25 strikethrough annotations into a table headed *"code-verified 2026-06-22"*, so the heading now over-claims. Added a note directly under it saying the date covers the unannotated cells only, and that struck cells carry their own date inline.

**Refused — "remove the strikethrough `1,571 tests green` text."** Striking it through is the fix; deleting it is not. That row is the evidence cell of a section whose entire purpose is to record what was true on 2026-06-22. Erasing the number destroys the record rather than correcting it — and this routine's scope rule is explicit that a stale number inside a dated record is *correct for its date*, so rewriting it falsifies the record. The strikethrough keeps the history readable while making unmissable that it is not current, and points at the measure-it command instead.

**Refused — "update `code-verified 2026-06-22` to the current audit date."** This one would have been actively false. That heading covers the `~90%` / `~30%` / `~5%` completion percentages and their evidence. I did not re-audit those; I corrected two stale cells. Stamping today's date on the section would assert a re-verification that nobody performed — manufacturing exactly the kind of untrue claim this PR exists to remove. I've written that reasoning into the doc itself, so the next reader (human or bot) doesn't retry it.

Worth naming the general pattern, since it will recur on every run of this nightly routine: an automated reviewer can't distinguish **"struck through as historical"** from **"still asserted"**, or **"dated record"** from **"stale claim"**. Both distinctions are load-bearing for how LIVING docs are maintained here. A future session that takes those summary lines at face value will delete history and forge a verification date. The in-doc note is there to stop that.

`doc_sync_check.py` green after the change (exit 0).

---

PR #424, opened 2026-08-25
fix(doc-sync): a guard is only as wide as what you feed it

Ranjith36963:
Cycle 15 measured the hole: `LIVING_DOCS` has **15** entries while **46** docs carry a `<!-- doc: LIVING -->` stamp. `documented_routes_exist()` and `collected_baseline_claims()` were both correct and both blind — their **input** was narrower than the estate they judged.

That is the third appearance of one shape today:

| Guard | Logic | Width |
|---|---|---|
| stamp presence | fine | 400-char window missed stamps below long front matter |
| `disagree:LOCATIONS` | fine | could only see disagreement, never five docs agreeing on a false number |
| routes / suite-baseline | fine | swept 15 files out of 46 |

The logic gets reviewed. The width never does.

## What changed

Both structural guards now sweep `living_stamped_docs()` — every file that says of itself that it is LIVING, and is therefore claiming to be checkable.

`LIVING_DOCS` stays for the **countable** guards: a number in an unexpected file is usually a quotation, not a claim. `constant_disagreements()` stays on the curated list for the same reason.

The suite-baseline regex also only ever watched `N collected`, so `DEPLOY.md:35`'s "1608 tests green" missed it **twice over** — wrong words *and* outside the swept set. Widened to accept `tests green` / `tests passing`.

## It earned immediately

Together the two widenings caught `MONETIZATION_GAPS.md:23` ("1,571 tests green") — which **cycle 15 itself did not find**. Neither widening alone would have seen it. Replaced with the measure-it command rather than a fresh unverifiable number.

## Performance, learned the hard way

The first cut used `ROOT.rglob("*.md")` and filtered `node_modules` out afterwards. **rglob walks a directory before the filter can reject it**, and this runs in the *blocking* CI step — it went from under a second to over two minutes.

It asks `git ls-files` now, which never descends there and agrees with `unstamped_docs()` about what "tracked" means, so the two guards cannot disagree about the estate. Memoised; three guards call it.

## Verified

- `doc_sync_check.py` → the only remaining rows are the two #423 fixes
- `doc_sync_mutation_test.py` → **All 25 guards proven able to go RED**
- `agent-gate.sh` → PASS

## Summary by CodeRabbit

* **Documentation**
  * Updated product-readiness guidance to reference the current collected test count rather than a fixed figure.
  * Improved documentation checks to keep living documentation and test-status statements aligned with the latest project state.

* **Bug Fixes**
  * Documentation validation now recognises additional wording for successful test results, reducing false warnings during checks.

### DIFF
--- a/docs/product/MONETIZATION_GAPS.md
+++ b/docs/product/MONETIZATION_GAPS.md
@@ -20,7 +20,7 @@
 
 | Dimension | % | Evidence |
 |---|---|---|
-| Product (features) | ~90% | engine + API + UI + auth + profiles + 4 scoring engines + control surface; 1,571 tests green |
+| Product (features) | ~90% | engine + API + UI + auth + profiles + 4 scoring engines + control surface; test suite green — measure the count with `python -m pytest --collect-only -q | tail -1`, never quote it |
 | Launch readiness (ship it live) | ~30% | no CI gate, no Dockerfile/deploy, no prod Redis+ARQ, no SES |
 | **Money layer (charge for it)** | **~5%** | **no Stripe, no billing, no plans, no pricing page, no analytics** |
 
--- a/scripts/doc_sync_check.py
+++ b/scripts/doc_sync_check.py
@@ -482,6 +482,64 @@ def landing_page_source_claims() -> list[tuple[int, int]]:
 
 
 DOC_KINDS = ("LIVING", "PLAN", "LOG", "REFERENCE", "FROZEN")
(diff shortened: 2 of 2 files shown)


Ranjith36963:
Superseded by #426, which carries these two commits rebased onto main after #423 and #425 landed (updating this branch in place needs a force-push, which this session is not permitted to do). #426 also adds the deletion contract the widening was built for.

PR #433, opened 2026-08-25
feat(docs): generate the countable facts instead of guarding them

Ranjith36963:
feat(docs): generate the countable facts instead of guarding them

The owner asked whether the doc system got stronger at CHECKING docs, at
READING truth out of code, and at WRITING docs. The first two: yes, 8 guards
to 26, all mutation-proven. The third: no. Nothing had changed about how docs
get written. This is that gap.

Seventeen nightly runs and ~2,260 lines of checker exist to notice drift in
facts that are DERIVABLE — the registry size, the rate-limit count, the
migration head, how many entries LOCATIONS holds. A guard makes drift VISIBLE.
A generator makes it IMPOSSIBLE, and needs no maintenance afterwards, because
there is no second copy to keep in step. Fable 5 ranked this below the deletion
contract and was right about the order; it was still the largest thing left
undone.

scripts/gen_doc_blocks.py fills marked regions in LIVING docs from the code:

       ...overwritten...   

`--write` fills them; the bare command CHECKS and exits 1 if the file on disk
disagrees. That check is now a BLOCKING step in doc-sync.yml — the only step in
that workflow that prevents wrong text rather than reporting it after the fact.
It reuses doc_sync_check.py's extractors rather than re-deriving anything: two
extractors for one fact is the duplication being removed, and if they ever
disagreed the guard and the generated text would fight forever.

Six facts in ARCHITECTURE.md are now generated: SOURCE_REGISTRY keys, unique
source classes, RATE_LIMITS entries, LOCATIONS entries, migration head,
SCORER_VERSION. The prose that restated them is deleted — a generated block
beside a hand-written restatement is two copies again, which is the problem
the block exists to end.

Proven both ways it can fail: a wrong value in the block is caught, and so is
DELETING the marker (a block quietly going back to being hand-written).

One design point worth keeping. living_surface() now EXCLUDES generated
regions, and swallows the trailing newline so a block costs exactly zero
surface lines. The surface metric counts prose that CAN become false; a block
written from code every build cannot. Counting it would punish the one fix that
ends drift — adding a generated table would raise the surface and trip the
ratchet, so nobody would ever do it. Excluded, converting prose into a
generated block makes the number FALL. The metric now rewards the right move.
Caught because the ratchet fired on this very commit.

Surface unchanged at 8,313: the six generated facts and their marker cost
nothing, and the restatements they replaced are gone.

Verified: gate PASS, 25 guards all proven able to go RED, ruff clean, YAML
parses, generated blocks current.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01D6DsQH5EUxJc3Vnqz9YHgQ

## Summary by CodeRabbit

* **Documentation**
  * Updated the architecture overview with current system statistics and clearer location, keyword, configuration and directory details.
  * Added automatic generation and refreshing of marked documentation sections.
  * Generated documentation now reflects the latest system facts.

* **Bug Fixes**
  * Improved documentation coverage accuracy by excluding generated content.
  * Added checks for outdated documentation, invalid directory references and unused configuration entries.

* **Chores**
  * Documentation validation now blocks changes when generated sections or mutation checks fail.


Ranjith36963:
Superseded by #434 — same work rebased onto main after #427/#431 landed, plus the guard/drill retirements that the rebase surfaced. Updating this branch in place needs a force-push, which this session is not permitted to do.

PR #445, opened 2026-08-26
Wire the journey: 16 places the product knew something and never acted on it

Ranjith36963:
**Draft — please read before merging. Merging `main` deploys straight to real users, and PR 3 below changes the email every user receives.**

Closes 16 of the 42 items in [`wiring.md`](../blob/feat/wire-pipeline-page/wiring.md), which maps every place the product knew something and never acted on it.

## The one-line version

The product was a good **filing cabinet** and not a **loop**. Everything before "Apply" was well built; the moment a user applied, the product went silent about the thing they now cared about most. The notification engine had **never read the `applications` table**.

## What's here, commit by commit

Each block is a separate commit and can be reviewed on its own.

| Commit | Block | What it fixes |
|---|---|---|
| `1b89491` | **The door** | The magic link dropped where you were going, so our own emails dead-ended on the dashboard. New accounts hit "adjust your filters" with no filters and no way forward. |
| `648a3e3` `8962bc5` | **The chase cron** | Nothing ever noticed "no reply". The 7-day query existed and was wired to one in-app banner. Now a daily cron speaks about your own quiet applications. |
| `cd39d49` | **The default email** | Instant mode (the default) sent a bare title and a link **straight to the employer**. It now carries score, verdict, reason and salary, and links back to us. |
| `ab4b94a` | **Don't lose the CV** | Regenerating a CV **destroyed** the one you applied with. Permanently. Now archived before the delete, and bound to the application. |
| `ca92427` | **Card truth** | The "Applied" filter returned zero for everyone, always. Cards looked live for dead jobs. Deadlines vanished on applying. |
| `4a5ca7a` | **Silent holes** | Clearing your profile left stale scores on screen. Three CV facts that move your score were invisible. |
| `22ab714` | **Launch gates** | No unsubscribe link existed at all. Four major actions emitted no analytics. |
| `53a6449` | **Drills** | `scripts/drill_wiring.py` fires the chase cron, the instant email, the pipeline board and unsubscribe on demand, and prints the real message a user would receive. |

## Two things that were destroying or misreporting data

- **`upsert_tailored_doc` was DELETE-then-INSERT.** Generate → apply → regenerate months later, and the file actually sent to the employer was gone from the database with no way to recover it. Migration `0033` archives before deleting.
- **`f"{title} at {company}".strip(" at")`** — `str.strip` takes a *set of characters*, not a suffix. "Data Engineer at **Meta**" went out as "at **Me**". Found by auditing my own work after shipping it; fixed in `8962bc5` with a guard over Meta / Monzo / Wise / Deloitte.

## Verification

Every item climbed rungs 1–4 of [`wiring_verification.md`](../blob/feat/wire-pipeline-page/wiring_verification.md):

1. **Red first** — or, where I wrote code first, retroactive **mutation testing**. Removing `safeNext`'s backslash guard failed exactly 2 of 21 tests; removing the chase cooldown failed 8; dropping the deadline failed exactly 1. Precision is the signal — a mutation that kills everything means coupled tests.
2. **Gate** green on every commit, real exit codes read.
3. **`api-types`** regenerated — and the drift guard caught a real break (`error TS2739`) before commit.
4. **Driven for real** — browser screenshots in `test-artifacts/rung4/` for the login journey, and the crons/emails run against real migrated Postgres, printing the actual subject and body.

### A security hole found in already-shipped code

`safeNext` accepted `/\evil.com`. It starts with `/` and not `//`, so the old check passed it — but browsers fold `\` into `/` per the WHATWG URL spec, making it `//evil.com` and off-site. **That was live in the password login path**, not just the magic-link path I was fixing.

## What this does NOT prove

**No real email has ever been sent.** No `RESEND_API_KEY` locally, and the Apprise boundary is stubbed in every test and drill. **Deliverability, spam placement and `List-Unsubscribe` are unproven** and can only be proven in production. The `List-Unsubscribe` header specifically is *not* available on this path — Apprise's mailto transport carries no custom headers — so the in-body link is what reaches recipients.

Also unproven: the shared dev database contradicts its own migration ledger (`applications` has no `user_id` though `0002_multi_tenant` is recorded as applied), so all verification ran against a clean database instead. And **two different migrations are both numbered `0031`** — `0031_universal_shelf` from another worktree and `0031_delivery_email_webhook_only` from main. That collision is already live and will bite whoever merges the other branch.

## Still open

26 items, including 7 product decisions now answered by the owner and queued as follow-up work: confirm-before-applied, "outreach" resetting the silence clock, user-confirmed ghosting, interview dates and reminders, a "your pipeline" digest section, and email click tracking.


Ranjith36963:
Superseded by a stack of 8 reviewable PRs, one per block — #447 → #454.

Same commits, same branch content, but each PR now shows ONLY its own block's diff instead of one 45-file change. Merge them bottom-up:

1. #447 — the map (docs only, merge-safe alone)
2. #448 — the door
3. #449 — the chase cron
4. #450 — the instant email
5. #451 — stop destroying the applied CV
6. #452 — card truth
7. #453 — silent holes
8. #454 — launch gates

Closing this one so there is no way to merge the same work twice.

PR #447, opened 2026-08-27
1/8 — The map: wiring.md + how each fix gets proved

Ranjith36963:
**Docs only. No product code.** Merge-safe on its own.

Two documents that the other seven PRs are measured against.

**`wiring.md`** — every hop in the journey where the product knew something and never acted on it. Found by two multi-agent sweeps over `origin/main` (26 agents). Every item carries a `file:line` for what exists **and the exact search that found nothing**.

It also carries its own honesty note: the adversarial refuters killed **0 of 37** findings, so these are high-confidence, *not* proven-by-fire.

**`wiring_verification.md`** — the five rungs each fix must climb, and the rule that matters: **"tests pass" is rung 2 of 5, not proof a user can use the thing.** Includes the per-item browser walkthroughs, the mutation results, and the six anti-patterns that have already cost this repo real time.

**Read this one first** — the other seven PRs only make sense against it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QpxEMPz2ZWG9Qed5BsFHvg

PR #448, opened 2026-08-27
2/8 — The door: the magic link keeps its destination, day one has a way forward

Ranjith36963:
Stacked on #447.

## W-01 — our own emails dead-ended at our own front door

The emailed sign-in link carried only `?token=`. The password form read a `next` destination; the **magic form — the default — never did**. So when the digest linked someone to `/jobs/123` and their session had expired, they signed in and landed on the dashboard. The job was gone.

`next` now rides the whole chain: `/login` → request → emailed link → consume → redirect. `safeNext` moved into `@/lib/safe-next` so both sign-in paths and the landing page share one definition.

### Security, two parts

**The link is built server-side**, so `next` is a server input. A browser-only check is bypassed by POSTing the API directly, and an unvalidated value there is an **open redirect carrying our own From: address**. `safe_next_path()` re-validates in Python, and the email builder calls it internally so no caller can route around the guard.

**It also closes a hole that was already live in the shipped password path:** `/\evil.com` starts with `/` and not `//`, so the old check passed it — but browsers fold the backslash into a forward slash per the WHATWG URL spec, making it `//evil.com` and off-site.

## W-03 — a new account hit a wall on minute one

Zero profile meant *"Try adjusting your filters, expanding the time range, or lowering the minimum score."* They have no filters, no search has ever run, and there is no score to lower. Nothing linked to `/profile`.

Now: **"Upload your CV"** with a working button. `hasProfile` is deliberately **three-state** — `undefined` while the profile query is in flight — so a returning user never gets that flash on every page load.

## Verified

Driven in a real browser: expired session → gated page → login → magic link → **landed back on `/pipeline`, not `/dashboard`**. Four hostile redirects posted **straight to the API**, bypassing the browser guard entirely — all dropped, control value kept. Screenshots in `test-artifacts/rung4/`.

Mutation-tested: removing the backslash guard failed **exactly 2 of 21** tests; collapsing the three-state check failed **exactly 1**.


Ranjith36963:

PR #449, opened 2026-08-27
3/8 — The chase cron: the product finally speaks about your own applications

Ranjith36963:
Stacked on #448.

## THE break in the whole map

Before this, **every message Job360 could send was about a job the user had NOT applied to.** The moment they applied, the product went silent about the thing they now cared about most — so "outreach", "no reply" and "ghosted" only existed if they remembered to open the Pipeline page and look.

That is a filing cabinet, not a loop.

## Everything it needed already existed and was wired to nothing

`get_stale_applications` — dormant 7+ days, correctly excluding offer/rejected — had **exactly two references in the entire backend**: its own definition and one route feeding an in-app banner. No cron, ever. And `grep applications` across `services/delivery` + `services/notifications` returned **zero**: the notification engine had never read that table.

## What lands

- **Migration 0032** — `applications.last_chased_at` plus an index. Without a cooldown, a daily cron re-chases every dormant row every day: an application quiet for 30 days would produce ~23 emails.
- **`get_applications_to_chase`** — deliberately *not* `get_stale_applications`. That one answers "what looks stale?" for a banner, which is free to repeat itself. This answers **"who should we INTERRUPT?"**, which is not idempotent.
- **`mark_application_chased`** — stamps `last_chased_at` and deliberately does **not** touch `updated_at`. That column means "the user did something"; us emailing is not that. Bumping it would reset the dormancy clock and the chase would silence the very signal it reports.
- **A daily 09:00 UTC cron**, registered in `WorkerSettings`.

## Deliberate choices, each with a test

- `match_score=None` — bypasses the score gate. *"Is this job good enough to mention?"* is the wrong question about a job they already applied to.
- `force=False` — quiet hours still apply. A chase is never worth waking someone at 3am.
- **Cap of 5 per user.** Someone who applied to 200 jobs then went on holiday gets one useful message, not 200. Leftovers stay eligible next run because only what was actually sent gets stamped.
- **No channel → cooldown not burned**, so a user who adds email a week later still hears about that first week.

## A bug I shipped and then caught

`f"{title} at {company}".strip(" at")` — `str.strip` takes a **set of characters**, not a suffix. **"Data Engineer at Meta" went out as "at Me".** Every employer ending in `a` or `t` was mangled. The 15 tests passed because I had used "Northwind" and "Acme" — **the failure was in my test data, not my test logic.** Fixed here with a guard over Meta / Monzo / Wise / Deloitte.

## Verified

15 tests, then **mutation-tested** because they passed first try: removing the cooldown killed 8; removing the `ghosted` exclusion killed exactly 1. Driven against **real migrated Postgres**. A drill in `scripts/drill_wiring.py` fires it on demand and prints the real message.

**Known bound:** chases are not written to `notification_ledger` — that table is `UNIQUE(user_id, job_id, channel)` and a chase is by definition about a job already in it. So chases do not yet appear in the in-app notification history.


Ranjith36963:

PR #450, opened 2026-08-27
4/8 — The instant email finally says what the dashboard says, and links back to us

Ranjith36963:
Stacked on #449.

**This one changes the email every user receives. Read it before merging.**

Instant is the **default** notify mode, and it was the worst message the product could produce. One line did two kinds of damage:

```python
body = f"Job360 match: {title}\n{apply_url}"
```

## W-17 — no score, no verdict, no reason, no salary

Everything that answers *"why am I being told about this?"* was missing — while the **digest path already had all of it** via `build_decision_card`. Two users on different settings received different products, and the default one was the poor one.

## W-18 — the link went straight to the employer

The digest links to `/jobs/{id}`. So in the default mode **the click could never come back**: no attribution, no staleness guard, no page that can still say "this one closed".

## The fix

`send_notification` now runs the **same** `user_feed` + `job_enrichment` join `send_bundle` uses, for one job, and renders through `build_decision_card`. `_card_lines()` is extracted as the **single definition of what a job says**, so the two modes cannot drift apart again — and there is a test that fails if they ever do.

**Before:** `Job360 match: Staff Site Reliability Engineer` + a raw employer link.
**After:** subject `Job360 — Staff Site Reliability Engineer at Monzo (88/100)`, body with London, £95k–£115k, `Fit 88/100 (strong fit)`, the reason, and `/jobs/7`.

## A third bug fell out of the same query

`match_score` — the input to dispatch's score-threshold gate — was read from a `SELECT` that **never fetched that column**. It was always `None`, so that gate has never gated anything for instant notifications. It now passes the score the user is actually ranked by.

## One deliberate behaviour change, checked before making it

The join is **INNER** on `user_feed`, so a job with no feed row is not sent — the rule the digest already states: we can neither score nor explain such a job, and an unexplainable alert is exactly the spam this product exists not to be.

Verified safe first: both enqueue sites are feed-driven (`main.py:509` enqueues *for feed rows*; `tasks.py:229` fires straight after `upsert_feed_row`), so production always has the row. Four existing tests encoded the old contract and were **updated to seed a feed row, rather than weakening the code to suit them**.

## Verified

7 tests **watched fail first** — the failure line was literally `assert '82' in 'Job360 match: Platform Engineer\nhttps://boards.example.com/...'`, both bugs visible at once. 136 passed across the notification, dispatcher, decision-card, digest-body and worker suites. Driven against real migrated Postgres with the actual subject and body printed and checked, 9/9.

**Bound:** no real email left the building — the Apprise boundary is stubbed. Deliverability is unproven and needs production.

### DIFF
--- a/backend/src/services/delivery/email_body.py
+++ b/backend/src/services/delivery/email_body.py
@@ -46,25 +46,69 @@ def render_digest_subject(*, shown: int, considered: int) -> str:
     return f"Job360 — {_plural(shown, 'job')} worth a look"
 
 
-def _render_card(card: DecisionCard, index: int) -> str:
-    """One job, as a small block. Numbered so a reply can name it later."""
-    lines = [f"{index}. {card.title} — {card.company}"]
+def _card_lines(card: DecisionCard, *, indent: str) -> list[str]:
+    """What one job SAYS — the single definition both send modes render from.
+
+    Split out of ``_render_card`` for wiring.md W-17: the instant path used to
+    hand-roll its own body (``"Job360 match: {title}\\n{apply_url}"``), so the
+    default notify mode shipped no score, no verdict, no reason and a link
+    straight to the employer, while the digest said all four. Two users on
+    different settings received different products.
+
+    Everything after the heading line is indented by ``indent`` so the digest can
+    keep its numbered layout while a single-job email reads flat.
+    """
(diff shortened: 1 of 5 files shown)


Ranjith36963:

PR #451, opened 2026-08-27
5/8 — Stop destroying the CV the user actually applied with (live data loss)

Ranjith36963:
Stacked on #450.

**This is live data loss, not a missing feature. It is happening in production right now.**

`tailored_documents` is `UNIQUE(user_id, job_id, doc_kind)` and `upsert_tailored_doc` is a **DELETE followed by an INSERT**. So there is no v1 and v2 — only ever "the current one".

Generate a CV → apply → regenerate months later, and **the file that actually went to the employer is gone from the database permanently**, with no way to recover it or even prove it existed.

Nothing tied an application to a document either, so *"which CV did I send for this job?"* could only be answered with *"whatever happens to be there now"* — a different question, and the wrong one.

## Migration 0033

One append-only table (`tailored_document_versions`) and two nullable columns on `applications`. Three hooks:

- **`upsert_tailored_doc` snapshots before the delete** — so a regenerate can no longer destroy anything. Taken *outside* the transaction on purpose: the archive must survive even if the regenerate itself fails.
- **`create_application` binds** whatever documents were in hand. That is the answer to the question.
- **`keep_tailored_doc` binds too**, because the common real order is *apply first, tailor and download afterwards*. Binding only at apply time would miss those people entirely. **First binding wins**, so a later rewrite cannot claim to be what was sent.

It stores the **text**, not a foreign key. The entire purpose is surviving the deletion of that row, and a pointer to a deleted row answers nothing.

## What it deliberately does not change

The live-document contract is untouched — `tailored_documents` still holds exactly one current row per (user, job, kind), so **not one existing reader changed**. Versioning that table instead would have meant widening the UNIQUE constraint every reader depends on, for no visible gain.

Applying with nothing tailored yet stores `NULL` rather than guessing. That honestly means "no document existed at that moment".

## Verified

13 tests **watched fail first** (*"regenerating destroyed the old document with no archive"*). 102 passed across tailoring, application, pipeline, chase and database suites. **Migration 0033 up AND down both execute on real Postgres with existing rows intact** — that needed checking because `test_migrations.py` builds a *fake* migrations directory, so no test in the suite ever runs the real files.

Walked on the real schema — tailor → apply → regenerate:

```
THE QUESTION: which CV did I apply with?
ANSWER: 'CV VERSION ONE — this is the file that went to the employer' (source=applied)
```

…while the live document correctly reads VERSION TWO.

Also includes `scripts/drill_wiring.py` — fires the chase cron, the instant email, the board and unsubscribe on demand, and prints the real message a user would receive. **Watched go red**: re-introducing the truncation bug made it print *"You applied to Platform Engineer at **Me**"* and fail.

### DIFF
--- a/backend/migrations/0033_tailored_document_versions.down.sql
+++ b/backend/migrations/0033_tailored_document_versions.down.sql
@@ -1,0 +1,5 @@
+-- Reverse 0033. Drop the columns first, then the table they point into.
+ALTER TABLE applications DROP COLUMN cover_letter_version_id;
+ALTER TABLE applications DROP COLUMN cv_version_id;
+DROP INDEX IF EXISTS idx_tailored_versions_user_job;
+DROP TABLE IF EXISTS tailored_document_versions;
--- a/backend/migrations/0033_tailored_document_versions.up.sql
+++ b/backend/migrations/0033_tailored_document_versions.up.sql
@@ -1,0 +1,49 @@
+-- 0033 — stop destroying the CV the user actually applied with (wiring.md W-08 + W-10).
+--
+-- Two failures, one table.
+--
+-- W-10: `tailored_documents` is UNIQUE(user_id, job_id, doc_kind) and
+--       `upsert_tailored_doc` is a DELETE followed by an INSERT. So there is no v1
+--       and v2 — only ever "the current one". Generate, apply, regenerate later, and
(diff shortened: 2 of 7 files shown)


Ranjith36963:

PR #452, opened 2026-08-27
6/8 — Card truth: the Applied filter works, dead jobs are flagged, deadlines survive

Ranjith36963:
Stacked on #451.

Three things the pipeline board was telling users that were not true.

## W-05 — the "Applied" filter returned an empty list for every user, always

The job list's "My Actions → Applied" filter read `user_actions`. The Apply button writes `applications`. So it returned **zero results for everyone, forever**, while looking like a working feature.

Fixed by **deriving, not duplicating**. The obvious fix — write `'applied'` into `user_actions` — would have been wrong: that table is one row per (user, job) with `ON CONFLICT DO UPDATE`, so it would **silently erase a `liked`** and the heart would vanish from the card. Two facts, two homes — so `applied` is its own field, and a job can honestly be both.

## W-15 — a card never learned its job had closed

Staleness was checked **once**, when the row was created, and never again. A card kept looking live for a job that closed months ago.

`expired` is true **only** for `confirmed_expired`, which comes from a direct URL check. `possibly_stale`/`likely_stale` are inferred from absence, and telling someone to stop chasing a live job is a worse error than staying quiet.

## W-16 (deadline half) — "Apply by 30 June" vanished the moment you applied

The date was never lost from the database; the pipeline query simply did not select it and the model had nowhere to put it.

**The interview half is deliberately untouched** — `interview_dates` stays a dead column here. (The owner has since decided to build it; that is follow-up work, not this PR.)

## Also wires a function written for this exact board and never called

`get_tailored_summary_for_jobs` — its docstring says *"for the Kanban attach"* and it had **zero callers**, leaving the CV/Letter button unable to say whether a document already existed. Called once per board, not once per card.

## A real constraint found while testing, not a test detail

For a logged-in caller `GET /api/jobs` reads `get_user_feed_jobs` (`routes/jobs.py:564`), **not** the shared catalog. A job with no `user_feed` row is invisible to the list no matter what else is true of it — so the Applied filter cannot show an application whose feed row has been purged. Written into `wiring.md` as a bound on the feature.

This surfaced because a **control** test — the `liked` filter, which should already have passed — was failing. Eight failures looked like "the feature isn't built yet"; the ninth meant something completely different.

## Verified

9 tests **watched fail first**, plus 2 more for the deadline half that were **mutation-tested** because they went green on the first run — forcing `deadline=None` killed exactly 1 and left the other 10 green. 115 passed across pipeline, API, IDOR, chase and document suites. `api-types` regenerated. A drill fires the whole board against real Postgres.

### DIFF
--- a/backend/scripts/drill_wiring.py
+++ b/backend/scripts/drill_wiring.py
@@ -244,7 +244,72 @@ async def drill_instant(db: Any) -> list[str]:
     return failures
 
 
-DRILLS = {"chase": drill_chase, "instant": drill_instant}
+async def drill_pipeline_card(db: Any) -> list[str]:
+    """W-05/W-15/W-16 — the board must tell the truth about each application."""
+    print("\n=== DRILL: pipeline card truth (W-05, W-15, W-16) ===")
+    failures: list[str] = []
+    jdb = JobDatabase.from_connection("", db)
+    await _seed_user(db)
+    job_id = await _seed_job(db, "Backend Engineer", "Wise")
+    await db.execute(
+        "UPDATE jobs SET deadline = ? WHERE id = ?", ("2026-09-30", job_id)
+    )
+    await db.execute(
+        "INSERT INTO applications (user_id, job_id, stage, created_at, updated_at) "
+        "VALUES (?, ?, 'applied', ?, ?)",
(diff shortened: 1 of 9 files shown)


Ranjith36963:

PR #453, opened 2026-08-27
7/8 — Silent holes: stale scores, an invisible score input, an untracked exit

Ranjith36963:
Stacked on #452.

Four places where the data was already right and something between the database and the user simply did not carry it. None is a missing feature; each is an unconnected wire.

## W-28 — the feed showed scores from a profile you had deleted

Uploading a CV or editing preferences queues a re-score. **Clearing a profile section or restoring an older version did not** — so the feed kept showing scores computed from a profile the user had just deleted or abandoned, presented as current. *An app showing a stale number as if it were fresh is an app lying.*

Called **bare** at both new sites, matching the two that already existed: `_maybe_trigger_rescore` documents its own guarantee (*"the profile save never 500s because of this"*) and falls back to an in-process task when Redis is down. Wrapping it again would hide that documented behaviour behind a second, different one.

The test breaks the **real enqueue door** rather than stubbing the function — my first version replaced the guard with my own and proved nothing.

## W-29 — three facts that move your score, invisible to you

`career_domain`, `cv_languages` and `cv_education_details` are read straight into the LLM matcher's prompt (`llm_matcher.py:249-260`), so they **move the score**, and the user could never see them, let alone correct a wrong one. Same "stored but not shown" class the `cv_positions` comment already documents.

Now on `CVDetail`, in `api-types`, and rendered on the profile under **"Used for matching"** with a line explaining they affect scoring and how to fix them. Each stays silent when empty (rule #29).

## W-12 — a tailored CV can now say it is out of date

Every document is stamped with the `profile_version` that produced it, and the response model dropped the field, so no *"written from an older version of your profile"* warning could ever be built.

`profile_changed_since` is set **only when both versions are known**. An unknown version is not evidence of staleness, and warning on it would train the user to ignore the warning. A *newer* version is not flagged either — that is a bug elsewhere, and calling it "out of date" would be actively wrong.

## W-06 — fixed as a copy change, deliberately

The job page had a second link saying *"view the complete listing and apply"* that wrote nothing, so anyone who used it applied without Job360 knowing.

The obvious fix — make it create an application — would have **manufactured applications for everyone who only wanted to read the description**, which is a worse lie than the one it replaces. So it now reads **"Read the full listing"**, the paragraph points at Apply, and exactly one control on the page claims to be applying.

## Verified

10 tests, watched fail first. **Three of the failures were my own setup being dishonest rather than the code being wrong**, and each was fixed rather than worked around: the fixture user has no profile (so `/profile/clear` correctly 404s), `UserProfile` holds `cv_data` not `cv`, and `cv_detail` is `None` without `raw_text` — which is right, because those three fields only exist *because* a CV was parsed.

210 passed. **The api-types drift guard immediately caught a real break** — adding three required fields to `CVDetail` invalidated the `EMPTY_CV_DETAIL` literal in `profile/page.tsx` (`error TS2739`). Fixed before commit; that is exactly the silent frontend/backend disagreement rung 3 exists to prevent.

### DIFF
--- a/backend/src/api/models.py
+++ b/backend/src/api/models.py
@@ -218,6 +218,14 @@ class CVDetail(BaseModel):
     # never SEE their parsed experience — only a "Roles: N" count. Part of
     # the "stored but not shown" gap closed 2026-08-08.
     cv_positions: list[dict[str, Any]] = []
+    # W-29 — these THREE are NOT display-only. They are read straight into the LLM
+    # matcher's prompt (services/llm_matcher.py:249-260), so they move the score.
+    # They were stored and never exposed, which meant a user could be ranked on a
+    # career domain or a language list they had no way to see, let alone correct.
+    # Same "stored but not shown" class as cv_positions above.
+    career_domain: str = ""
+    cv_languages: list[str] = []
+    cv_education_details: list[str] = []
     # Projects stated on the CV. A Projects heading was already used as a
     # section boundary and then discarded — for a junior or career-changing
     # candidate it is often the strongest evidence on the document.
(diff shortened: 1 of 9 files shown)


Ranjith36963:

PR #454, opened 2026-08-27
8/8 — Launch gates: unsubscribe, analytics, and the first deletions

Ranjith36963:
Stacked on #453. Last in the stack.

## W-23 — there was no unsubscribe link. At all.

The only way to stop the emails was to log in and delete the channel. That is a **deliverability problem before it is a legal one**: a recipient who cannot find the exit presses "spam" instead, and that is the worst signal a sending domain can collect.

**Stateless token, no table:** `{user_id}.{hmac_sha256(SESSION_SECRET, user_id)[:32]}`. No migration, no cleanup job, no row that can drift out of sync with the account it names. Not guessable without the secret; revocable in bulk by rotating it.

**Safe to expose unauthenticated because of what it can do:** the only outcome is silence. It cannot read anything, change an address, or reach an account. A leaked token buys an attacker the ability to *stop someone's email*, which the owner reverses by logging in.

**Deliberately never expires** — an unsubscribe link must work in a year-old email.

**POST behind a confirm button, never a GET.** Email clients and security scanners *prefetch* links, so a state-changing GET would unsubscribe people who never clicked. Same trap `/auth/magic` already solves, solved the same way.

Unsubscribing sets `notification_rules.enabled = 0` and **deletes nothing** — channels, quiet hours and mode all survive, so turning it back on restores exactly what they had rather than an empty default. The confirmation page says how to undo it.

**Known bound, in the code:** Apprise's mailto transport carries no custom headers, so a `List-Unsubscribe` header is **not available on this path**. The in-body link is what reaches recipients.

## W-27 — the funnel went dark right after Apply

Four events, each fired when the action **succeeds**, never on the attempt — an event on intent counts moves that failed, which is how a funnel quietly starts lying about itself:

`application_stage_changed` · `tailored_document_generated` · `tailored_document_downloaded` (with `doc_kind` **and** `format` — a bare count cannot tell a CV from a cover letter) · `channel_added`

## Delete sweep — partial, deliberately

- **D-03 `getActionCounts`** — zero references anywhere. Deleted.
- **D-02 `keepTailored`** — the dead frontend client is deleted. **The backend route is kept**: it is exercised by `test_cv_coverletter.py`, it is public API surface, and PR 5 made `keep_tailored_doc` load-bearing for document binding. Deleting a tested route to tidy up is riskier than the note proposing it assumed, and earns nothing today.

`tsc --noEmit` passing *after* the deletions is the actual proof they were dead — grep tells you what it can see, the type checker tells you what the program needs.

## Also carries the owner's seven decisions

The final commit records all seven product calls as spec, including the two that went **against** my recommendation (build interview tracking; add email click tracking) with the trade-off stated once so it is not re-argued later.

## Verified

14 unsubscribe tests — **the forgery cases are the ones that matter**: a token that silences an account must not be guessable, must not accept a swapped user id, and must fail closed when an email client truncates the URL. 61 passed across the email, worker and chase suites. A fourth drill fires the real unsubscribe path against real Postgres and prints the exact line that goes into every email.

### DIFF
--- a/backend/scripts/drill_wiring.py
+++ b/backend/scripts/drill_wiring.py
@@ -305,10 +305,64 @@ async def drill_pipeline_card(db: Any) -> list[str]:
     return failures
 
 
+async def drill_unsubscribe(db: Any) -> list[str]:
+    """W-23 — the exit in every email must actually work."""
+    print("\n=== DRILL: unsubscribe (W-23) ===")
+    failures: list[str] = []
+    from src.services.notifications.unsubscribe import (
+        unsubscribe_line,
+        verify_token,
+    )
+
+    await _seed_user(db)
+    await db.execute(
+        "INSERT INTO notification_rules (user_id, enabled) VALUES (?, 1)", (DRILL_USER,)
+    )
+    await db.commit()
+
+    line = unsubscribe_line(SITE_BASE_URL, DRILL_USER)
+    print("  the line that goes in every email:")
+    print("   ", line)
+    token = line.rsplit("token=", 1)[1]
+
+    jdb = JobDatabase.from_connection("", db)
+    await jdb.set_notifications_enabled(verify_token(token) or "", enabled=False)
+    cur = await db.execute(
+        "SELECT enabled FROM notification_rules WHERE user_id = ?", (DRILL_USER,)
+    )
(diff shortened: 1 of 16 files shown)


Ranjith36963:

PR #455, opened 2026-08-28
docs: daily truth check

Ranjith36963:
Nightly scout pass. Every claim below was verified against **code**, never against another doc. Findings already reported by `scripts/doc_sync_check.py` are excluded as already-handled.

**LIVING surface: 10,291 → 10,285 lines** (57 docs). `doc_sync_check`: **53 → 50** drifted claims. `doc_sync_mutation_test`: **24/24** guards still prove-able.

---

### 1. `.claude/skills/sync/SKILL.md` — closed by DELETE + POINTER
**Doc said** the `LIVING_DOCS` constant is "the only authority and has grown from 6 files to **15**", and told the nightly agent to stamp that list.
**Code says** `LIVING_DOCS` holds **19** entries while **57** tracked docs carry `<!-- doc: LIVING -->`. The structural guards deliberately do *not* read it — `doc_sync_check.living_stamped_docs` takes every tracked `.md` at its own word, and its docstring states outright that `LIVING_DOCS` "stays for the COUNTABLE guards" because it is narrower.
**Why it mattered:** an agent obeying this stamped 19 of 57 docs and left 38 to go red on freshness. Today's guard run shows exactly that failure in progress. Replaced the count and the wrong authority with a pointer to `doc_sync_check.living_stamped_docs`.

### 2. `.claude/skills/verify-job360/SKILL.md` — a **fixed bug documented as live**; closed by DELETE
**Doc said** (as a hard-won gotcha): "the web app saves profiles per-user … but `run_search` loads `load_profile(DEFAULT_TENANT_ID)`. So a logged-in user's 'New Search' runs profile-less."
**Code says** `src.main.run_search` loads `load_profile(user_id or DEFAULT_TENANT_ID)`, and `api.routes.search` passes `user_id=user.id`. The fix is pinned by an existing regression test, `tests/test_main_scheduler_wiring.py::test_run_search_loads_profile_for_given_user`, and the source comment names the same report this gotcha came from.
**Why it mattered:** this is an instruction executed on every verification run, sending the verifier to chase a bug that was fixed. Deleted outright — closure (3) was already in place.

### 3. `.claude/skills/verify-job360/SKILL.md` — dead pointer; closed by DELETE
**Doc said** "See `E2E_TEST_REPORT.md` at the repo root for the format and the two real bugs this methodology already caught."
**Code says** no such file is tracked anywhere in the repo.

### 4. `.claude/skills/verify-job360/SKILL.md` — stale count; closed by DELETE
**Doc said** `test_main.py` has "**14** E2E tests" running "in ~8s".
**Code says** 18 test functions, no parametrisation. Deleted the count and the timing; kept the load-bearing instruction (*do not re-add `--ignore=tests/test_main.py`*).

### 5. All **415** raw `file.py:NN` citations in LIVING docs — closed by POINTER (path kept, rotting number dropped)
These had rotted **wholesale**, silently, exactly as the ratchet's own docstring predicts. Measured examples:

| Doc claim | What that line actually is | Real location |
|---|---|---|
| `settings.py:227` = `EMBED_BACKFILL_PER_RUN` | `MIN_TITLE_GATE` | 267 |
| `settings.py:221` = `SEMANTIC_ENABLED` | a "gate-pass scoring" comment | 261 |
| `settings.py:151/152/155` = the `ENRICHMENT_*` trio | comments about the *storage* floor | 191/192/195 |
| `settings.py:115` = `MIN_STORE_SCORE` | "Delivery is email + webhook only" | 160 |
| `settings.py:60/61` = `OPENAI_API_KEY`/`OPENAI_MODEL` | `REED_API_KEY` / `ADZUNA_APP_ID` | 88/89 |
| `settings.py:279-336` = `RATE_LIMITS` | a comment | 383-440 |
| `core/settings.py:387-400` = `validate_required_env` | `RATE_LIMITS` dict entries | ~497 |
| `main.py:1137-1163` = the enrichment gate | the `INSERT OR IGNORE` block | ~1247 |
| `main.py:853`, `main.py:1137`, `rescore.py:85`, `tasks.py:237` = the Engine-2 gate call sites | — | 910, 1222, 174/432/614, 243 |

Only **one** (`settings.py:25` = `DATABASE_URL`) was still correct. The transform is provably minimal — a verifier confirmed nothing but `:NN` suffixes changed: no `()` call-parens, no `` `:NN` `` continuation refs, no ports, no line counts. The enrichment-gate sentence in `02-search-and-match-engine.md` was additionally repointed at `src.main.run_search` and `workers.tasks.enrich_job_task`.

---

## Outside the stated edit scope — please read

`scripts/line_citation_baseline.txt` is **not** a `.md`, but leaving it would have silently **disabled a CI guard**. The ratchet compares per-file counts against that baseline; with every count now 0 but the old ceilings (e.g. `02-search-and-match-engine.md 70`) still recorded, a doc could have absorbed 70 new raw line numbers without the guard firing. The mutation test caught this and went RED, which is the file's own documented trigger for re-baselining ("These may only ever go DOWN").

Re-baselined to **empty** — the *tightest* setting, not a disabled one: `line_citation_regressions` reads a missing entry as `base.get(rel, 0)`, so **any** raw line number added to **any** LIVING doc now fails immediately. Mutation test passes again, 24/24.

## Step 4 — closing the loop

No new checks proposed for `doc_sync_check.py`; per the brief, that script should stop growing, and the empty ratchet now covers finding 5 permanently. Two observations for a future cycle instead:

- **The existing `skill_dead_paths` guard has a blind spot.** It missed both finding 3 (`E2E_TEST_REPORT.md`) and a bare `api/profile.py` reference in `worker`/`integrator` SKILL.md — the real module is `api/routes/profile.py`. Widening that guard's *input* (rather than adding a new guard) would catch this class. This is the same "a guard is only as wide as what you feed it" shape that `living_stamped_docs` was written to fix.
- **Finding 2 needed no new test** — `test_run_search_loads_profile_for_given_user` already existed. The doc simply never learned. Worth noting that closure (3) only holds if the doc is deleted when the test lands.

## Not fixed (reported, not touched)

- `frontend/tests/design/README.md` carries no `<!-- doc: KIND -->` stamp. Already reported by the guard; per the brief I do not guess a doc's kind.
- The 50 remaining `doc_sync_check` findings are the guard's own and were left to it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YTXQotLDfdHVJZW9Ae5ZGY)_


Ranjith36963:
**Conflict resolved, but I could not publish the result — the branch update needs a hand.**

## The conflict is fixed

`main` moved 15 commits (the bring/tailor/receipt/MCP pivot). Both conflicts were `main` restructuring sections this PR had only *mechanically* transformed, so both were taken from `main` wholesale and the citation strip re-applied on top:

- `.claude/skills/hard-rules/SKILL.md` — main reordered the rule sections
- `docs/product/pillars/01-user-pillar.md` — main rewrote it around the pivot

All four content findings were **re-verified against current `main`** and still hold. Finding 1 got *worse*, not stale: `LIVING_DOCS` still holds 19 while the stamped estate grew 57 → 61, so `/sync` now misses 42 docs instead of 38.

Verified on the rebuilt tree: `doc_sync_check` **66 → 63**, mutation test **24/24**, LIVING surface **10,775 → 10,769**, raw line citations **415 → 2**.

## What is blocking

I rebuilt the branch onto current `main` as one docs-only commit rather than a merge commit, for a concrete reason: `.claude/hooks/commit-gate.sh` waives the gate only when a commit touches **nothing** under `backend/` or `frontend/`. A merge commit stages main's backend changes, so it demands a full-suite stamp — and the container this ran in has no Postgres, no `psycopg` and no `pytest`, so no honest stamp was possible. I did not fabricate one or bypass the hook.

Publishing that rebuilt branch needs a **force-push**, which was declined by the permission policy in this session. So `origin/docs/daily-truth-check` is still at the old `da8d747` and remains conflicted. **Nothing here is lost** — the finished commit exists locally and as a patch that applies cleanly to `1fba085`.

To land it, either re-run this with force-push allowed, or apply the patch onto a fresh branch off current `main` and point a new PR at it.

## Two things found along the way, neither fixed here

1. **`scripts/agent-gate.sh` is broken on bash 5.2** — it aborts before running any test, so *no* agent on Ubuntu 24.04 can produce a gate stamp. Line 61 puts `|| true` on the line that opens a heredoc inside a command substitution:
   ```bash
   LOCK_OUT="$(cd backend 2>/dev/null && python - <<'PYLOCK' 2>/dev/null || true
   ```
   `bash -n` passes (the static parse defers), but at runtime bash 5.2 re-parses and dies with ``syntax error near unexpected token `||` ``. Minimal repro:
   ```bash
   X="$(cd /tmp && python3 - <<'PY' 2>/dev/null || true
   print("HI")
   PY
   )"
   ```
   Fix is to move `|| true` off the heredoc-opening line (e.g. close the substitution first, then `|| true` on the assignment). I left it alone: it is the commit gate, it is outside this PR's docs-only scope, and I cannot run the suite here to prove a change to it.

2. **`backend/CLAUDE.md` keeps its 2 citations** — it sits under `backend/`, so touching it trips the same gate requirement. Rather than leave them uncounted, I pinned that file at `2` in `line_citation_baseline.txt` with a note; a 3rd still fails the ratchet. Any session that can run the suite should strip them and delete the row.

---

Ranjith36963:
Follow-up to the comment above: the rebuilt work is now on the remote, so it no longer depends on a local container or the patch file.

**Branch: [`docs/daily-truth-check-rebuilt`](https://github.com/Ranjith36963/job360/tree/docs/daily-truth-check-rebuilt)** @ `19261ec` — exactly **1 commit ahead of `main`, 0 behind**, so it is conflict-free by construction.

I pushed to a new ref rather than updating this PR's branch because updating it needs a force-push (the branch was rebuilt onto current `main`, not merged), and that was declined by the permission policy. This PR's branch is deliberately untouched at `da8d747` and still conflicted.

To land it, either open a PR from `docs/daily-truth-check-rebuilt` and close this one, or re-run with force-push allowed and I'll move this PR's branch across. I have not opened a second PR — that is your call, not mine to make unilaterally.

---

Ranjith36963:
Closing: this daily truth-check is CONFLICTING against main and edits docs that were deleted whole on 2026-09-05 (docs/product/pillars/*, STORY.md, etc.). Superseded by the later checks and by #510.

PR #467, opened 2026-09-02
feat(hygiene): give the broom an entrance — auto-reap shipped worktrees + branches

Ranjith36963:
## The finding

Two correct tools already solve the hard part of worktree sprawl. Neither has ever collected anything.

| tool | what it does | why it never ran |
|---|---|---|
| `worktree_census.py` | sorts SAFE / KEEP / ASK, handles squash-merge, sees gitignored work | manual command |
| `branch_prune.sh` | deletes merged branches, real safety contract | manual command **and** asks the wrong oracle |

Between them being written and today the estate went **14 → 62 → 98** worktrees.

> **A cleanup tool nobody runs is indistinguishable from no cleanup tool.**

## Why `branch_prune.sh` found nothing for five weeks

```
git branch --merged origin/main   ->    1 branch
gh pr list --state merged         ->  343 branches
```

The repo **squash-merges**. A squash rewrites the commits, so a shipped branch is never an ancestor of `main`. Every stock "delete merged branches" recipe reads the first number, correctly concludes there is nothing to do, and exits 0. It was not broken — it was reading a number that is always 1.

`worktree_census.py` already computes the right answer with `git cherry` patch-ids. Nothing acted on it.

## What this adds

**No classifier.** It imports the census and trusts its verdicts. Four additions:

1. **An entrance** — `.claude/hooks/worktree-reaper.sh` at SessionStart. It only *launches*: `gh` is 2.16 s, one `worktree remove` is 10–30 s, a census over 30 worktrees measured **3 m 25 s**. Inline that is a several-minute session start, so it forks detached and reports what the *last* run did.
2. **An idle floor** — the census asks whether a folder holds *work*. ~14 sessions run in parallel here; a clean, shipped worktree somebody is simply not typing in passes every test it has. Mtime is the only cheap evidence nobody is home. An unreadable mtime reads as ACTIVE, never as stale.
3. **Prune first** — a merged PR auto-deletes the branch, breaking the worktree's `.git` link, and then `git worktree remove` **refuses outright**. Only `prune` clears those; 17 of 68 were in that state. Pinned by a drill against a real temp repo, not by a comment.
4. **Branch deletion on the right oracle** — acting on the census's patch-id verdict instead of `--merged`.

## Measured on the live estate

- **0 worktrees** reapable — the census is stricter than a naive dirty-check, because it sees gitignored `.env` and `backend/data/` that `git status` prints *nothing* for.
- **170 local branches** squash-merged and safe.

Hence **two separate caps**. A worktree is a recursive rmdir; a branch is a ref write. One shared cap would have drained 170 branches at 10 per 6-hour run — four months for work that takes a second.

## Drills — 20 cases, on every CI push

Registered in `drill_registry.py`, and added to `ci.yml` rather than only the weekly workflow: a break in the one component whose job is *deleting things* should not sit undetected for seven days.

Every group carries a **negative control**. A reaper that refuses everything passes every safety case and is exactly as useless as no reaper — which is how `branch_prune.sh` looked correct the whole time.

`no_second_classifier` parses this file's own AST and fails if any `run([...])` reaches for `status` / `cherry` / `merge-base` / `--merged`. Two brains disagreeing about SAFE is worse than one being wrong, because the disagreement stays invisible until it deletes something. Watched red by planting exactly that call.

## Recovery

Every tip SHA goes to `docs/maintenance/REAPED.log` **before** deletion. Reflog is not an undo: `git branch -d` deletes the branch's own reflog with it, and HEAD reflogs are per-worktree.

Kill switch: `touch docs/maintenance/REAPER-OFF`.

`branch-reaper.yml` deliberately does **not** commit that log — a remote branch's merged PR is a better record than any file, and pushing weekly to `main` would mean a weekly production deploy to write down what GitHub already stores.

### DIFF
--- a/.claude/hooks/worktree-reaper.sh
+++ b/.claude/hooks/worktree-reaper.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# worktree-reaper.sh — SessionStart hook. LAUNCHES the reaper; never IS the reaper.
+#
+# WHY THIS EXISTS
+#   Nothing tears a worktree down when its PR lands, so they reached 98.
+#   `scripts/branch_prune.sh` was written for exactly this in July 2026 and is
+#   wired into no workflow and no hook, which is the same as not existing. This
+#   file is the entrance it never got.
+#
+# WHY IT ONLY LAUNCHES
+#   A SessionStart hook sits on the critical path of every session opening, and
+#   the work does not fit that budget. Measured on this repo:
+#       gh pr list ................ 2.16 s
+#       git worktree list ......... 0.43 s across 30 worktrees
+#       one `worktree remove` ..... 10-30 s (each carries node_modules + .venv)
+#   So this forks a detached child and returns. Inline, a session start would pay
+#   several minutes.
+#
+# WHY THERE IS ALMOST NO LOGIC HERE
+#   .claude/hooks/ is NOT scanned by scripts/drill_registry.py, so anything
+#   decided in this file is decided somewhere nothing can drill. A throttle that
+#   silently never fires looks identical to one that works. So the throttle, the
+#   report and every safety gate live in worktree_reaper.py, which has 20 drill
+#   cases; this file only resolves paths and forks.
+#
+# CONTRACT
+#   - FAILS OPEN. Any error exits 0. A cleanup hook must never stop a session.
+#   - THROTTLED (--hook-tick, default 6 h). ~14 sessions open against this repo;
+#     unthrottled, that is 14 concurrent `gh` calls, i.e. a rate limit not a sweep.
+#   - CAPPED (--max 3). It converges over a few sessions instead of blocking one.
+#
+# KILL SWITCH
+#   touch docs/maintenance/REAPER-OFF   (the child re-checks it too)
+set -uo pipefail
+trap 'exit 0' ERR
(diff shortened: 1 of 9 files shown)


Ranjith36963:
⚠️ **This merge broke production.** The branch was stale — based on a main from before slices 1-2 of the pivot — and merging it reverted newer work wholesale (4,401 deletions across 33 files in the product paths alone):

- `backend/src/api/routes/{oauth,well_known,receipts,tokens}.py`, `backend/src/services/auth/{oauth_flow,oauth_clients,api_tokens}.py` — the entire OAuth 2.1 server (#488, merged yesterday) — **gone from main**
- `frontend/src/app/{oauth,bring,receipts,settings/connect}` pages, middleware protected paths, `api.ts` helpers, the OAuth block in `.env.example` — gone
- Live impact, verified: `https://job360.uk/.well-known/oauth-authorization-server` → **404** (was serving the discovery doc yesterday); Railway backend is running `2278e9c`.

Fix in flight: revert PR from branch `revert/pr-467-clobber` (full revert of `2278e9c`), gate running now. **Fastest owner action: click "Revert" on this PR and merge, or merge my revert PR when it opens.** The reaper feature itself can be re-landed later from a fresh branch cut off current main.

PR #468, opened 2026-09-02
fix(security): clear the two high-severity advisories blocking every merge since 31 Aug

Ranjith36963:
`npm audit (frontend deps)` is one of the **eleven required contexts** in `main-production-gate`. It has been failing on **`main` itself since 2026-08-31T10:11:58Z** — so no PR in this repo can merge, by hand or through the queue. Every other auto-merge fix is worthless while this is red.

## The advisories

| package | advisories |
|---|---|
| **browserslist** | `GHSA-c83g-rgw3-j3cx` unbounded memory growth → OOM · `GHSA-73wf-gq98-2v4g` crash / prototype write |
| **fast-uri** | `GHSA-5jgf-p345-68v8` host confusion · `GHSA-f65p-4m7j-42xc` SSRF via IPv6 · `GHSA-fph4-wmhf-6fwf` SSRF via percent-decoding · `GHSA-jqff-g426-hqxp` host confusion |

Both transitive. Nothing here imports either directly.

## The fix

`npm audit fix --package-lock-only` — **lockfile only.** `frontend/package.json` is untouched, so no declared range moved and nothing is a semver judgement.

```
before: 5 vulnerabilities (1 low, 2 moderate, 2 high)
after:  found 0 vulnerabilities
```

Lockfile re-validated: parses, lockfileVersion 3, 1021 packages.

**Kept deliberately alone** — it unblocks every other PR, so it should be reviewable in ten seconds and mergeable on its own rather than carried in with a harness change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FEs1dix6f5imaMkuxuAAgd

PR #470, opened 2026-09-03
docs: daily truth check

Ranjith36963:
Scout pass. **Rebuilt on current `main` (`962d88e`) — this PR is much smaller than it was, and the reason matters.**

## What happened to the original PR

The sourcing era was deleted while this sat open (#499). **97 files left `backend/src`** — `main.py`, `skill_matcher.py`, `rescore.py`, `workers/tasks.py` among them — and the four pillar docs moved to `docs/_archive/sourcing-era/` stamped **FROZEN**. The LIVING surface fell **10,656 → 4,970** lines.

Everything this branch previously carried was about that code and those docs:

| original finding | status now |
|---|---|
| Stale Pillar-2 engine-gate line citations in nine LIVING docs | **dropped** — the cited modules are deleted and five of those docs are now FROZEN archives |
| `runbook.md` account-removal recipe (`strftime`, reversible soft-delete) | **dropped** — `runbook.md` is a FROZEN archive |
| `verify-job360` skill's `datetime('now')` UPDATE | **dropped** — already gone from `main` |
| `verify-job360` skill's `E2E_TEST_REPORT.md` pointer | **still true** — fixed here |

I did not merge the old branch forward. Resolving those conflicts would have written my edits **into FROZEN archives** — falsifying a record of a moment — and re-pointed citations at modules that no longer exist. So the branch was rebuilt from `main` and only what is still true was re-applied.

## The two findings that survive — both agent-executed instructions, both DELETE

**1. `.claude/skills/verify-job360/SKILL.md`** sent the reader to `` `E2E_TEST_REPORT.md` at the repo root `` for the report format and "the two real bugs this methodology already caught". That file **has never existed in this repository** — `git log --all -- E2E_TEST_REPORT.md` is empty, so it is not a deletion that outran its reference; it never landed.

**2. `.claude/skills/worker/SKILL.md` paragraph 5a** repealed a hands-off rule over `src/services/{job_enrichment,llm_matcher,scoring_dimensions,skill_matcher,embeddings,retrieval,vector_index}.py`, cited `skill_matcher.py:587` and the guard `tests/test_scorer.py::test_dims_neutral_not_zero_when_enrichment_missing`, and told the agent that rules #18/#19/#20/#27 still bind. Checked one by one: **all seven modules gone, the guard test gone, and `.claude/skills/hard-rules/SKILL.md` — which held those numbered rules — gone too.** A repeal notice for files that no longer exist cannot bind anything, and nothing load-bearing is lost by removing it (there is no Pillar-2 work left to refuse). Deleted rather than reworded.

These two files are read and executed by agents nightly, which is why a dead reference in them costs more than one in prose.

`scripts/line_citation_baseline.txt`: `worker/SKILL.md` **1 → 0** — the deletion took its only raw line citation with it.

## Verification

- **LIVING surface 4,970 → 4,968.**
- `python scripts/doc_sync_check.py` — **5 findings before, 5 after.** All five are `main`'s own: workflow count (24 vs 25), endpoint count (80 vs 84), test-file count (148 vs 156), a `backend/data` dead tree path, and the surface ceiling `main` already exceeds (4,970 vs 4,951). None introduced here, none fixed here.
- `python scripts/doc_sync_mutation_test.py` — **10/10 guards still able to go red.** That set shrank from 24 with the sourcing era; measured on this tree rather than quoted from the earlier run.
- Both removed strings re-checked absent from every non-archive doc afterwards.

## Note on the force-push

This branch contained only my own commits. It was reset onto `origin/main` and pushed with `--force-with-lease` pinned to the previous head, because the old content could not be honestly merged forward.

## Still open, for a human

`scripts/agent-gate.sh` remains unrunnable on bash 5.2 — a `|| true` on a here-doc's opening line inside a command substitution, so `bash -n` passes but the runtime re-parse fails under `set -euo pipefail`. One-line fix, verified, in a comment below. Separately, its commit-gate hook counts `git merge --abort` as commit-creating and blocks it, though that command un-does a merge rather than making one.

https://claude.ai/code/session_019k4v5YHeJWGuKe5t7HLyRu


Ranjith36963:
## `npm audit (frontend deps)` is red here, and it is not this PR's

**The check:** `npm audit --omit=dev --audit-level=high` → exit 1, two high-severity advisories, both transitive:

- `browserslist <=4.28.6` — GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g
- `fast-uri 3.0.0 - 3.1.5` — GHSA-5jgf-p345-68v8, GHSA-f65p-4m7j-42xc, GHSA-fph4-wmhf-6fwf, GHSA-jqff-g426-hqxp

**Why it isn't this PR's:** the diff is 9 markdown files plus one `.txt` ratchet baseline — zero npm files. And the same check is red on the base branch: `security.yml`'s scheduled run on `main` @ `e58cceb` ([run 33381308371](https://github.com/Ranjith36963/job360/actions/runs/33381308371)) failed.

**The fix already exists: #468.** It is fully green — `npm audit` included — its merge-base with `origin/main` is `9724d47` (the same commit this branch is based on), and it touches exactly one file, `frontend/package-lock.json` (browserslist 4.28.2→4.28.8, fast-uri 3.1.5→3.1.7 plus transitives).

I verified that fix rather than assuming it, running both lockfiles locally in `frontend/`:

```
main's lockfile   → npm audit --omit=dev --audit-level=high → exit 1, "2 high severity vulnerabilities"
#468's lockfile   → same command                            → exit 0, "found 0 vulnerabilities"
```

## Why I did not push the port into this PR

I prepared the port — `frontend/package-lock.json` taken verbatim from #468, verified byte-identical to its own diff — and then could not commit it, for a reason worth knowing on its own:

**`scripts/agent-gate.sh` cannot run on bash 5.2.** It dies at the Postgres singleton-lock probe before any test:

```
scripts/agent-gate.sh: command substitution: line 104: syntax error near unexpected token `||'
```

The cause is the `|| true` sitting on the here-doc's *opening* line inside a command substitution (`scripts/agent-gate.sh`, the `LOCK_OUT=` assignment). `bash -n` passes at file level, so no syntax guard catches it — but bash re-parses `$( … )` content at runtime, and there the trailing `|| true)` lands after the here-doc body and fails to parse. Under `set -euo pipefail` that kills the gate. Minimal repro:

```bash
set -euo pipefail
X="$(true && cat - <<'PY' 2>/dev/null || true
hello
PY
)"
```

Moving `|| true` to *after* the closing paren fixes it — verified:

```bash
X="$(true && cat - <<'PY' 2>/dev/null
hello
PY
)" || true          # → GOT:[hello], exit 0
```

That is a one-line change in shared tooling that every session on this platform depends on, so I have not made it here — it does not belong in a docs truth-check PR, and it is the owner's call.

Independently, this container could not have produced a green gate anyway: no `pytest`, no Postgres on :5433, and no `frontend/node_modules` for the `test:unit` / `type-check` / `lint` leg. So the honest position is: the fix is validated and ready, but pushing it from here would mean either bypassing a legitimate commit gate or editing shared tooling unilaterally. Neither is mine to do.

**What unblocks this PR:** merge #468. `npm audit` then goes green here on the next run with no change to this branch. I have not re-run the failed check, since a re-run cannot pass against an unchanged lockfile.

---

Ranjith36963:
## Correcting my last comment: don't merge #468 — merge `main` in, which I've now done

My earlier comment said "merge #468 and this goes green." That advice is now wrong twice over, and the situation is better than it looked.

**1. `main` is already green.** The lockfile moved with #459 (`9724d47..97e6fbc`, 90+/101− in `frontend/package-lock.json`). Run on Linux against current `main`:

```
npm audit --omit=dev --audit-level=high  →  exit 0, "found 0 vulnerabilities"
```

So the merge freeze is over. This PR was simply behind — based on `9724d47`, the last commit with the red lockfile. I've merged `origin/main` into the branch (`89b0ba1`, no conflicts), so `npm audit` now runs against the fixed lockfile. No port of anything, no manifest guess from me.

**2. Closing #468 was right, and I can confirm the diagnosis you couldn't test.** This container is Linux with node 22 / npm 10.9.7, so I ran the check your machine couldn't:

| lockfile | `npm ci` on Linux |
|---|---|
| #468 `2c9509d` (your Windows `npm audit fix`) | **fails**, verbatim your error: `Missing: @emnapi/runtime@1.11.3 from lock file` · `Invalid: lock file's @emnapi/wasi-threads@1.2.2 does not satisfy 1.2.3` |
| #468 `afcbb83` (pr-repair's follow-up) | **succeeds** — and the lockfile is byte-identical afterwards (sha256 unchanged), then `npm audit --omit=dev --audit-level=high` exits 0 |

Your platform diagnosis was exactly right for the commit you wrote. Worth recording that pr-repair's follow-up commit had already repaired it and met the blind checker's stated bar in full (*"`npm ci` must succeed and leave the lockfile byte-identical, then `npm audit` must exit 0"*) — that verdict stayed UNCERTAIN only because that agent had no network either. Moot now that `main` carries a fix, but it means the branch you closed was not the broken thing by the end.

## Re-verified after the merge

`main` brought in new backend and frontend code, so I re-checked this PR's own claims against the moved tree rather than assuming:

- **All ten symbol pointers still resolve** — `rescore.rescore_user_feed`, `skill_matcher.JobScorer.score`, `main.run_search`, `main._run_matcher_stage`, `workers.tasks.score_and_ingest`, `cv_parser._maybe_normalise_skills_via_esco`, `JobDatabase.hard_delete_user` / `soft_delete_user`, and the cited test. This is the point of the change: the line numbers they replaced would have rotted again on this very merge.
- **`doc_sync_check`, measured in two real worktrees rather than eyeballed:**

  | | LIVING lines | drifted claims |
  |---|---|---|
  | `origin/main` @ `97e6fbc` | 10,423 | 63 |
  | this branch @ `89b0ba1` | **10,420** | **63** |

  −3 lines, **zero** new findings. (Both numbers rose against my original PR body's 10,288/54 because `main` gained code and two new LIVING docs — not because of anything here.)

## Still open, unchanged

`scripts/agent-gate.sh` remains unrunnable on bash 5.2 — it dies at the `LOCK_OUT=` probe before any test, because `|| true` sits on the here-doc's opening line inside a command substitution. Details and the one-line fix are in my previous comment. It didn't block this merge commit (the gate hook exempts a clean tree), but it will block any commit here that touches `backend/` or `frontend/`.

---

Ranjith36963:
## `Advise (never merges)` is red — the merge cage is broken on `main`, on every PR

`npm audit` went green on this head after the merge, as expected. A different check went red, and it is not this PR's either — it is a `main` regression that silently disables the cage repo-wide.

**The failure** ([run 33739103655](https://github.com/Ranjith36963/job360/actions/runs/33739103655/job/100596612674)):

```
File "scripts/merge_cage.py", line 479, in <module>
    ALLOW, DENY = _load_lists()
File "scripts/merge_cage.py", line 389, in _load_lists
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

**Why it isn't this PR's:**

- `.claude/`, `docs/` and one `.txt` are all this PR touches — nothing under `scripts/`, nothing in `.github/workflows/`.
- **It is red on #473 too** ([job 100591682067](https://github.com/Ranjith36963/job360/actions/runs/33737567262/job/100591682067)), an unrelated `feat/mcp-server` branch. Every PR gets this now.
- The `import yaml` arrived with **#444** (`36c1c84`, *"one list, two readers"*), which is on `main`. `.github/workflows/pr-advisor.yml` runs `python scripts/merge_cage.py --drill` with `actions/setup-python@v7` and **no install step between them** — so PyYAML was never on that runner. It ran at import, before argparse, which is why nothing downstream got a chance.

**Reproduced and fix verified locally**, in a clean venv running the workflow's exact command:

```
$ python -m venv /tmp/v            # no PyYAML, like the runner
$ /tmp/v/bin/python scripts/merge_cage.py --drill
  ... ModuleNotFoundError: No module named 'yaml'          → exit 1   (identical traceback)

$ /tmp/v/bin/pip install pyyaml
$ /tmp/v/bin/python scripts/merge_cage.py --drill
  COVERED 25/25 decision-path functions; 25 blockers on record, all drilled.
  DRILL RESULT: 101/101 passed                             → exit 0
```

**Proposed patch** — one step in `.github/workflows/pr-advisor.yml`, between `setup-python` and the drill:

```yaml
       - uses: actions/setup-python@v7
         with:
           python-version: "3.12"
 
+      - name: The cage needs its policy reader
+        run: pip install pyyaml
+
       - name: The cage still refuses — and can still allow
         run: python scripts/merge_cage.py --drill
```

I have **not** applied it here. It is a CI workflow change on shared tooling, unrelated to a docs truth check, and widening this PR to carry it is exactly what the drive-to-green rules tell me not to do. It also wants a second look at whether any *other* `merge_cage` / `lane.py` caller runs on a runner without PyYAML.

**Why this one is worth your attention beyond my PR:** the cage's own comment says a missing or malformed policy *"must never read as 'no restrictions' — that is the `else: allow` bug wearing a different hat."* The import failure lands in the same place from the other side: the drill dies, the advisory step then posts *"The merge cage could not judge this PR… the absence of a verdict here means nothing — it is not a clean bill of health."* That message is doing its job, but it means **no PR has had a real cage verdict since #444 landed** — and it reads as a bot hiccup rather than a guard being down.

I have not spent a re-run on this check: the failure is deterministic (reproduced locally, and red on an unrelated PR), so a re-run against unchanged code cannot pass.

---

Ranjith36963:
## Base recovered — the PyYAML fix landed, so I've merged `main` in

`main` moved to `7c6f433` and `.github/workflows/pr-advisor.yml` now installs PyYAML in the `advise` job, with the cause recorded in the comment ("*`lane.py` imports PyYAML, and since `36c1c84` so does `merge_cage.py`*"). That was the whole failure. Merged into this branch as `7bc8ad6` — no conflicts — so `Advise (never merges)` re-runs against a base that can actually load its policy.

To be explicit about what I did **not** do: I didn't apply that workflow patch myself, and this merge doesn't carry a hand-written version of it — it's `main`'s own fix arriving through the base.

## Re-verified after the second base move

`main` also brought `VISION.md` and edits to **`ARCHITECTURE.md` and `STATUS.md`** — two files this PR changes — so I re-checked rather than trusting a clean `git merge`:

- **Citation ratchet holds.** I lowered `ARCHITECTURE.md` to 38 and `STATUS.md` to 9 in this PR; after the merge they measure **exactly 38 and 9**. `main`'s new prose added no raw line citations, so nothing I tightened went red. `README.md` 6, `glossary.md` 6, `runbook.md` 22 — all at baseline.
- **The three `line-citation-ratchet` rows are pre-existing.** They fire on `origin/main` by itself (3 rows, the Universal Shelf pillar docs). Not introduced here — deliberately left unbaselined, per the reasoning in `scripts/line_citation_baseline.txt`.
- **All nine symbol pointers still resolve** against the moved tree. That is now twice across two independent base moves; every line number they replaced would have needed re-checking both times.
- **`doc_sync_check`, measured in two real worktrees:**

  | | LIVING lines | drifted claims |
  |---|---|---|
  | `origin/main` @ `7c6f433` | 10,656 | 66 |
  | this branch @ `7bc8ad6` | **10,653** | **66** |

  −3 lines, **zero** new findings — the same delta as against the previous base.

The absolute numbers keep climbing (10,288 → 10,420 → 10,656) purely because `main` keeps gaining LIVING docs. The delta is the part this PR controls, and it has been −3/0 at every base.

## One note on the automation

`finding-watch` stopped the auto-fixer after 3 attempts and said this one wanted human eyes — correct, since the failure was never fixable on this branch. `pr-shepherd` then re-dispatched the repair loop anyway, and `pr-repair` has failed to finish twice ([33738643141](https://github.com/Ranjith36963/job360/actions/runs/33738643141), [33739310435](https://github.com/Ranjith36963/job360/actions/runs/33739310435)) without pushing anything. Not blocking now that the base is fixed, but the two bots disagree about when to stop, and the shepherd's counter appears to win.

`scripts/agent-gate.sh` is still unrunnable on bash 5.2 — details and the one-line fix are in my earlier comment, unaffected by any of this.

---

Ranjith36963:
## `drift-check` fixed — and the attempt turned up a real flaw in this routine's own rules

**The failure** was `STALE GENERATED BLOCKS — run: python scripts/gen_doc_blocks.py --write` on `ARCHITECTURE.md`. Not this PR's: the branch hadn't touched that file, and `origin/main` @ `962d88e` **fails the identical check**. Since the remedy is the repo's own generator rather than a hand edit, porting it is deterministic — `a48dbc4`, one line inside the `<!-- generated -->` fence, workflow count 24 → 25. `gen_doc_blocks.py --check` now passes.

### The part worth your attention

While in `ARCHITECTURE.md` I tried to also close the three hand-written counts `doc_sync_check` reports — `CLAUDE.md` "24 workflows", `ARCHITECTURE.md` "80 endpoints" and "148 `test_*.py` files". Under the deletion contract this routine runs on, they look like the ideal target: each is a stale number sitting **right beside its own instruction not to quote it**. Verbatim:

> across **148** `test_*.py` files *(collected-test count: measure it, never quote it)*

> the GitHub Actions harness — **24** workflows in `.github/workflows/` *(…; measure with `ls`)*

I deleted all three. `doc_sync_check` then emitted **four new findings**:

```
| (all docs) | - | workflows     | claim not found in any doc | reworded/removed — update checker patterns or restore the claim |
| (all docs) | - | route-modules | claim not found in any doc | ...
| (all docs) | - | endpoints     | claim not found in any doc | ...
| (all docs) | - | test-files    | claim not found in any doc | ...
```

**The guard is right and the contract's general rule does not apply to these.** A countable fact the checker *watches* cannot rot silently — it is caught the day the code moves. Deleting the sentence doesn't retire the claim, it **blinds the check**. The deletion rule earns its keep against prose that nothing verifies; against a watched fact it is strictly worse than leaving the sentence alone.

So I reverted all three. They stay, and they stay the guard's finding to report rather than this PR's to fix — the routine's own instruction is that anything `doc_sync_check` reports is already handled.

I'm flagging it because the daily prompt tells the scout to DELETE restatements as the first choice, with no carve-out for facts the checker watches by name. A future cycle following it literally will keep re-discovering this the hard way — and each time will look, for one commit, like progress.

## State of this PR

| | |
|---|---|
| `gen_doc_blocks.py --check` | passes (was the failing check) |
| `doc_sync_check.py` | **5 findings before, 5 after** — the same five `main` carries |
| `doc_sync_check.py --budget-only` | passes |
| Mutation test | 10/10 guards able to go red |
| LIVING surface | **4,968** (from `main`'s 4,970) |

Still three files of substance: two dead references deleted from agent-executed skills, plus this regeneration.

---

Ranjith36963:
Closing: this daily truth-check is CONFLICTING against main and edits docs that were deleted whole on 2026-09-05 (docs/product/pillars/*, STORY.md, etc.). Superseded by the later checks and by #510.

Ranjith36963:
Closing makes sense — and I'm not reopening. One factual note for the record, checked against `main` @ `ca89364` just now:

- **`worker/SKILL.md` paragraph 5a — genuinely superseded.** Already gone from `main`. Nothing lost.
- **The `E2E_TEST_REPORT.md` pointer — still live.** `.claude/skills/verify-job360/SKILL.md:143` still says *"See `E2E_TEST_REPORT.md` at the repo root"*, and `git ls-tree` confirms no such file exists (nor ever has — `git log --all` on it is empty). One line, in a file agents execute nightly.

Flagging rather than acting: it's yours to decide whether that's worth a follow-up, and a later truth-check will find it again regardless.

The other thing from this branch worth not losing is `scripts/agent-gate.sh` on bash 5.2 — it dies at the `LOCK_OUT=` probe before running any test, so no commit touching `backend/` or `frontend/` can be gated from a bash 5.2 machine. Cause and the one-line fix are [in this comment](https://github.com/Ranjith36963/job360/pull/470#issuecomment-5518854179); unrelated to the docs, so it doesn't go away with this PR.

---

PR #566, opened 2026-09-14
docs: daily truth check

Ranjith36963:
Scout pass over the LIVING docs — the drift `scripts/doc_sync_check.py` and `scripts/doc_sync_mutation_test.py` cannot see. Both were run first and are green before and after this diff (`No drift`; `All 12 guards proven able to go RED`). Nothing they already pin is duplicated here.

**LIVING surface: 3,207 → 3,133 lines (−74).** 11 docs changed, 48 insertions / 117 deletions, no doc deleted or renamed, no code touched.

Every fix is a DELETE or a POINTER-by-symbol. No claim was reworded.

> **Note on the merge (`dfe4e67`) and the repair (`50d0b3a`).** `main` moved because **#565 landed — a parallel run of this same routine**. Merging it here resolved `ARCHITECTURE.md` and `.claude/skills/verify-job360/SKILL.md` to this branch's *older* side, silently reverting six of #565's fixes. Each was re-verified against code and restored in `50d0b3a`; see the [comment](https://github.com/Ranjith36963/job360/pull/566#issuecomment-5663539218) for the detail. Findings 6, 7 and parts of 1–3 below are therefore now carried by main rather than by this diff — they are kept in the list because they were independently found and verified here, and two runs converging on them is the corroboration.

## Findings

1. **`ARCHITECTURE.md`** — doc said "Auto-purge: `purge_old_jobs(days=30)` deletes jobs by liveness … skipping any row whose `source` is the user-brought marker (hard rule 3)", citing `repositories/database.py:688,723`. Code says there is **no `purge_old_jobs`** anywhere in the repo (only a mention inside `migrations/0010_run_log_observability.up.sql`'s comment), and **hard rule 3 does not exist** — `.claude/skills/hard-rules/SKILL.md` indexes 14 rules and 3 is not among them. Nothing purges jobs. *(#565 found this too.)* — **closed by DELETE**

2. **`ARCHITECTURE.md`** — doc restated the baseline `jobs` DDL, and its `CREATE TABLE jobs` omitted `match_score`, `visa_flag`, `experience_level` and `last_seen_at`, all of which `JobDatabase.init_db` actually creates. The `Pragmas:` paragraph cited `database.py:94` for a line that lives at `database.py:77`. — **closed by DELETE** (the 40-line SQL block and the three paragraphs restating it) **+ POINTER** to `repositories.database.JobDatabase.init_db`, `repositories.pg.translate`, `backend/migrations/`, and the existing `tests/test_pg_translate.py`

[...]

PR #560, opened 2026-09-12
docs: daily truth check

Ranjith36963:
> **Body rewritten 2026-09-14.** The original described a 40-doc, 3,644-line LIVING surface and a test CI had not yet run. None of that is true any more: #559 (slice 8) pruned the docs, #565 and #566 landed their own truth checks, and the owner merged `main` in and resolved the conflicts. Leaving the old description up would have been the exact failure this PR exists to fix. The original text is in this PR's edit history.

Scout pass over the LIVING docs — the drift `scripts/doc_sync_check.py` cannot see.

**LIVING surface: 32 docs, 3,127 lines.** Guards green before and after: `doc_sync_check.py` → "No drift"; `doc_sync_mutation_test.py` → "All 12 guards proven able to go RED".

`docs: updated` — 8 docs + 1 new test file. No source changes.

## What survives in this diff

The base moved a long way under this branch, and **#565 / #566 independently fixed roughly half my original findings** — `verify-job360/SKILL.md`, `ARCHITECTURE.md`, `troubleshooting.md`, `CONTRIBUTING.md`, `STATUS.md` and `reviewer-bugs.md` now match `main`, so they have dropped out of the diff entirely. Good outcome; three sessions converged on the same truths. What is left is what only this branch found:

| File | Doc said | Code says | Closed by |
|---|---|---|---|
| `docs/product/BREACH-RUNBOOK.md` | Hour 0–1 **step 4**: deliberate rotating `CHANNEL_ENCRYPTION_KEY`, because it makes "every stored notification-channel credential permanently unreadable — users will have to reconnect their channels" | No Python reads that variable; `services/channels/` was deleted 2026-09-05 and `0040` dropped `user_channels`. There is nothing to make unreadable | **DELETE** the step |
| `docs/product/BREACH-RUNBOOK.md` | Step 3 rotate list ends with "job-source API keys" | No job sources exist (slice 5, #483) | **DELETE**; replaced with "read the live names off Railway → Variables", which cannot rot |
| `docs/product/BREACH-RUNBOOK.md` | Sensitivity map lists `user_actions` as live job-hunt data | Dropped by `0040_drop_notification_tables` | **TEST** (below) |
| `docs/product/BREACH-RUNBOOK.md` | Two `<!-- doc: LIVING -->` stamps | One is enough | **DELETE** the duplicate |
[...]

PR #569, opened 2026-09-14
docs: daily truth check — the breach runbook's stale freshness footer

Ranjith36963:
Follow-up to #566. Opened as a three-finding PR against `docs/product/BREACH-RUNBOOK.md`; **#560 — a third parallel run of this routine — landed two of those three first, and did them better.** What is left is one finding.

**This PR is now a four-line diff: 1 insertion, 3 deletions, one file.** `scripts/doc_sync_check.py` green. LIVING surface **3,125** lines.

## The remaining finding

**Footer** — the doc ended with "Review this page every 6 months … **Last verified against the real stack: 2026-07-24** (sessions table + SESSION_SECRET dual kill-switch, **Fernet channel-cred rotation trade-off**, R2 ciphertext-only backups, Resend sender domain)." Two things wrong with it: the Fernet channel-credential mechanism it cites as evidence of freshness was deleted 2026-09-05 (no Python reads `CHANNEL_ENCRYPTION_KEY`, `cryptography` is not in `backend/pyproject.toml`), and the doc already carries a machine-checked `<!-- doc: LIVING | last-verified: … -->` stamp on line 1. A second, older, hand-typed freshness claim underneath it can only ever rot — and this one had. — **closed by DELETE**

## What #560 already fixed, better than my version did

Kept here for the record, since the original body claimed them:

- **The dead `CHANNEL_ENCRYPTION_KEY` containment step.** Both runs found it. #560's fix is on `main`.
- **Step 3's key list.** I trimmed "job-source API keys" off a hand-typed list of secret names; #560 deleted the list outright and pointed at the live Railway variable names instead. Deleting beats trimming — I have taken #560's version.
- **`user_actions` in the sensitivity map.** My original body said I deliberately did not chase this, because the doc's own "enumerate the live schema first" hedge seemed to cover it. **That reasoning was wrong** — the hedge tells you not to trust the table, but the table still named a dropped one, and an incident responder reading it would have believed a nonexistent table held their job-hunt activity. #560 verified it and pinned it with `backend/tests/test_dropped_tables_stay_dropped.py`. Confirmed independently here: `0040_drop_notification_tables` drops `user_actions` and nothing recreates it after — `0002_multi_tenant`'s drop-and-recreate is earlier, so the name is grep-visible in the tree while the table does not survive to head. That is exactly the trap, and a test is the right closure for it.

## Note on the merge

The merge `62c94d6` resolved `BREACH-RUNBOOK.md` to this branch's older side, reverting all four of #560's improvements — and #560 is already on `main`, so merging this PR at that commit would have regressed `main`. `4c72e7b` restores them; see [the comment](https://github.com/Ranjith36963/job360/pull/569#issuecomment-5663908106). Everything in this file except the footer is now byte-identical to `main`.

This is the second merge today (after #566's) whose conflict resolution silently reverted the newer fix. Both leave no trace in the final diff, and `doc_sync_check.py` stays green through them because the reverted claims are prose, not countable facts. **A WIP limit of one open `docs: daily truth check` at a time would remove the class** — three were open simultaneously today.

docs: updated docs/product/BREACH-RUNBOOK.md

https://claude.ai/code/session_01XppoTbQdNcpZXz2YDtghRP


Ranjith36963:
Second time today, same shape — flagging it because this one would have regressed `main`.

The merge `62c94d6` says "the newest truth-check wins on BREACH-RUNBOOK.md", but the resolution took **this branch's older side for the whole file**, reverting four of #560's fixes. #560 is already merged to `main`, so merging #569 as it stood would have undone them there. Pushed `4c72e7b` to restore.

What was being reverted:

- **`user_actions` back into the breach sensitivity map.** Verified against code, not against either doc: `0040_drop_notification_tables` drops it and nothing recreates it after (`0002_multi_tenant` drops-and-recreates as part of its rebuild, but 0040 is later and final). It does not exist at migration head. The map would have told an incident responder that a nonexistent table holds their job-hunt activity, during the assessment step where they are deciding what to report to the ICO.
- **The `test_dropped_tables_stay_dropped.py` citation**, replaced by "(migrations `0031`, `0039`, `0040`)" — a named standing test swapped back for migration numbers.
- **Step 3's hand-typed key list**, which #560 had deleted in favour of reading the live names off Railway Variables. My own version only trimmed that list; #560's deletion is the better fix and I've taken it.
- **The duplicate `<!-- doc: LIVING | Fable compliance finding … -->` stamp** on line 4.

Only the footer fix — dropping "Last verified against the real stack: 2026-07-24 (… Fernet channel-cred rotation trade-off …)", which cited the dead mechanism as evidence of freshness — survives as this PR's own change. Everything else here is now identical to `main`. **This PR is a four-line diff.**

Worth saying plainly: #560 did this file better than I did. It deleted the key list where I trimmed it, and it chased `user_actions` where I looked at the same ambiguous `DROP`/`CREATE` pair, decided the doc's own "enumerate the live schema first" hedge covered it, and moved on. The hedge did not cover it — the map still named a dropped table. #560 verified it and pinned it with a test, which is the right closure.

The pattern across #566 and #569: three runs of this routine were open at once, and in both merges the conflict resolution silently reverted the newer fix. It leaves no trace in the final diff, and `doc_sync_check.py` stays green throughout because the reverted claims are prose, not countable facts. A WIP limit of one open `docs: daily truth check` would remove the whole class.

---

